PR #21, opened 2015-03-11
Update 2-validating-domain-objects.rst

simonschaufi:
removed forgotten asterisk at the end of the Validator and fixed Copies of objects. Don't know what sidebar was used for.

### DIFF
--- a/Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
+++ b/Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
@@ -77,7 +77,6 @@ many validators for default requirements like the validation of emails,
 numbers or strings.
 
 
-
 When does validation take place?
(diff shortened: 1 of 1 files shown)


marble:
Let's keep the sidebar. It's for related thoughts that are not in the direct line of text. It's rendered as a separate box on the right. I've just given that a slightly better look (border+margin). Reload your browser. Can you please submit the pull request again without changing the sidebar? And yes, keep on helping - go, go, go : - ))

simonschaufi:
ok looks better but the general look is still somehow strange. done: #22

marble:
It's important to keep the right structure=semantics. Don't think in terms of layout. As soon as there's time we'll improve the html rendering. Ah, BTW, maybe I can setup PDF rendering for this book. Then it'll be a box again!

PR #23, opened 2015-03-22
Update 3-implement-individual-database-queries.rst

simonschaufi:
can syntax highlighting also be used with fenced blocks like:

```php
function bla() {
	$phpcode = $bla;
}
```

? or does the typo3 docs renderer not understand this?


marble:
Hi Simon, yes it's possible. See http://mbless.de/blog/2015/03/02/php-syntax-highlighting-in-sphinx.html for the background. Would you be so kind and add a short ticket at https://forge.typo3.org/projects/team-doc-rendering ? Thanks a lot!

simonschaufi:
done: https://forge.typo3.org/issues/65931

marble:
Thank you Simon!

PR #29, opened 2015-07-24
Add pluginNamespace option

pogosheep:
Add the option with short description.

It's implemented at `\TYPO3\CMS\Extbase\Service\ExtensionService->getPluginNamespace()`

### DIFF
--- a/Documentation/b-ExtbaseReference/Index.rst
+++ b/Documentation/b-ExtbaseReference/Index.rst
@@ -236,6 +236,11 @@ the following sections:
 	Therefore you need to copy all original templates to this folder before you set
 	this TypoScript setting.
 
+``view.pluginNamespace``
+	This can be used to specify an alternative namespace for the plugin.
+	Use this to shorten extabse daults plugin namespace,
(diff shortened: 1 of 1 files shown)


marble:
Thanks! I removed some typos, I think. I hope it's alright with you. http://docs.typo3.org/typo3cms/ExtbaseFluidBook/b-ExtbaseReference/Index.html#typoscript-configuration

pogosheep:
Yes thanks ... I've missed one thing ... take a look at http://docs.typo3.org/typo3cms/ExtbaseFluidBook/b-ExtbaseReference/Index.html#typoscript-configuration with updated rendering. The order is wrong. We should add `pluginNamespace` after `templateRootPath`

marble:
Oops, that was me, I changed that order. Because everything is sorted alphabetically there. Why would you change that?

pogosheep:
Because the other options are all related. You configure where to look for specific files like templates.

The new option is related to other things like GET-Parameter and such. But if sorting is alphabetical it's okay to me, didn't notice the sorting.

PR #34, opened 2015-09-23
A few corrections to the first paragraphs of 5-debugging-with-xdebug.rst

SoulCover:


marble:
Thank you very much! Don't hesitate to send more pull requests. Note that in:
:file:`something.php` there must not be a blank after the colon..

PR #35, opened 2015-10-05
update php code and use namespace

PeterTheOne:


marble:
Hello Peter, thanks for your contribution. But since the whole book uses the notation of classes without namespaces I rather choose to not check your patch in as it wouldn't be consistent with what we have at 1001 outher places in that book. But of course, we take you patch as reminder that an update of the whole book would be disirable.

PeterTheOne:
Thanks, I understand. I think even only updating all the namespaces would in itself go a long way.

PR #36, opened 2015-10-16
[TASK] Add missing documentation for acceptsEmptyValues feature

DanielSiepmann:
* As validators won't get called for empty values anymore, this was a
  breaking change and should be documented

Note: This was changed in some version. Perhaps we should annotate the version in some way.


marble:
As we are discussing already in the mailinglist: Let's wait with this PR until we have different versions

PR #38, opened 2015-10-28
Migrate Extbase class names

DanielSiepmann:
As extbase now uses namespaces, this should be reflected in documentation.


marble:
I guess you know what you are doing :-)

PR #40, opened 2015-11-01
[CLEANUP] Remove old code syntax

DanielSiepmann:
* Streamline inline code highlighting to use `` instead of :code:


marble:
Hello Daniel, seems we had a misunderstanding. Remove JUST the ':code:' part, which is the explicite naming of the textrole. The inline code itself stays <ONE backtick>this is the code<ONE backtick>. Two backticks mean 'literal text' which is partly the same - but not quite, as "code" has a styling of its own. So it's like writing 'code' (use backticks) in Slack and Markdown. Can you do that replacement again?
Meanwhile I'll add that `.. default-role:: code` directive to Includes.txt.

PR #42, opened 2015-11-02
[CLEANUP] Remove old code syntax

DanielSiepmann:
* Streamline inline code highlighting to use single back tics instead of
  :code: and single back ticks, as code is the default role
* Also use spaces instead of tabs in those files, to follow PSR2


DanielSiepmann:
I'll update the rest of indention while working on

PR #43, opened 2015-11-02
Feature/update b extbase reference

DanielSiepmann:
I recommend to ignore white spaces as all files I've worked on are reindented with 4 spaces instead of tabs to follow PSR2 as the TYPO3.CMS Core.

Only the Extbase reference was updated. All other files were modified to get more labels for linking from the reference.

i've tried to make most changes topic related in there own commits. Just the last commit is huge as it's the main work on the reference. So perhaps it's easier to proof the PR by checking each commit.


marble:
Tip: .. todo:: is a directive. If you use that, you can get a list of open todos rendered IN you text. With links and so on. Output needs to be turned on in Settings.cfg or Makefile, don't remember by heart. Search for site:sphinx-doc.org directive todo

marble:
Klasse!

PR #49, opened 2015-11-05
Kasper Skårhøy -> Kasper Skårhøj

neil-h:
At least Google seems to think Kasper's last name is spelled with a j :)

Also, I have no idea why it says the last line here is different. I didn't change anything there, just the spelling of Kasper's last name.


DanielSiepmann:
That can be an editor setting and change of line feed / new line.

Please not that we think about a rework / write of the book in most parts. So your work can be obsolete in the near future in case you fix more errors in later chapters.

Can I get in contact with you via email / chat or something else? Perhaps you are willing to proof read the changes / new work?

PR #55, opened 2015-11-13
[BUGFIX] Fixes markup for table name

funkvn:


marble:
Thank you! Gimme more :-))

PR #68, opened 2016-03-15
Changes field name to l10n_*

funkvn:
L18n makes no sense: localisation or internationalization, you have to decide. :-)


DanielSiepmann:
Yes, but ... poorly that are the current "conventions" in TYPO3 CMS. Even if that makes it harder for new developers and is not "official".
See core extension for frontend: https://git.typo3.org/Packages/TYPO3.CMS.git/blob/HEAD:/typo3/sysext/frontend/ext_tables.sql#l297

But of course you can add a hint to the documentation that this is based on history and should be treated like l10n.

Perhaps even the one thought it will handle i18n and L10n and decided on that base.

But we can ask the core team if that doesn't satisfy you.

I'll leave the Pull Request open for further discussions.

marble:
Yes, `l10n` would be better. I suggest we use that in the documentation as soon as it has changed in the real (code) world.

PR #72, opened 2016-04-29
Adjusted a CREATE STATEMENT and PHP-Code

AMartinNo1:
The CREATE STATEMENT for the relation tx_inventory_domain_model_product ends with KEY PARENT (pid), ); The last comma must not be there.

Despite that it is a good habit not to close the PHP.tag unless it is necessary to avoid the possible bug "headers already sent".


marble:
Thank you very much. Don't hesitate to send more!

PR #76, opened 2016-06-01
Update to namespaces and fix form.textbox

vertexvaar:
Introduce namespaced FQCNs and replace non existent f:form.textbox with f:form.textfield


DanielSiepmann:
Thanks for the improvements @vertexvaar.

PR #79, opened 2016-06-01
Correctly indent example TCA PHP (persistence/configure..backends..input..

vertexvaar:
Original link: https://docs.typo3.org/typo3cms/ExtbaseFluidBook/6-Persistence/2-configure-the-backends-inputforms.html


DanielSiepmann:
I'm not fully sure whether the indentation is right now.
Did you test the rendered version?

If not, I'll try this afternoon. As tabs are used and at least github will use 8 spaces per tab. The rendered result should be 4 spaces per indent according to PSR-2.

DanielSiepmann:
Checked against working example above, looks like our sphinx rendering will render it properly.

DanielSiepmann:
The result looks like it should. Thanks.

PR #86, opened 2016-06-29
Update 4-and-action.rst

edrush:
The update() function does not work, which seems to be an open issue for long time. Let's not make others spend time on that but tell them directly.


marble:
Yes, it is good to have that information. On the other hand: Just to have the information "does currently (?) not work" doesn't really tell much in the end. Can you open a bug tracker ticker somewhere? Then the information WOULD make sense, as we can add a link to that issue too. And can have hopen that it will somewhen somehow be fixed. What do you think?

edrush:
Right, thank you @marble. I was searching for a while on https://forge.typo3.org, but gave up, today I created an issue.

marble:
:-))

PR #88, opened 2016-07-14
Update 5-using-typoscript-for-rendering-the-cobject-viewhelper.rst

fischhase:
look at:
https://docs.typo3.org/typo3cms/TyposcriptReference/Gifbuilder/Gifbuilder/Index.html


marble:
Thanks! From looking at the docs I would say: Yeah, you're right!

PR #90, opened 2016-07-19
add comment about performance / caching

macjohnny:


macjohnny:
backport to 7.6

marble:
Hi Esteban, many thanks for your contributions. There's no branch 7.6. We "only" have 4.5, which is the original book, und "latest" which is thought of "with namespaces, >6.x".

Note that reST has a ".. node::" directive. And we have some more textroles, like for files and php. I've applied that and as a result we get:
image

Have a look at the textroles here: https://docs.typo3.org/typo3cms/drafts/github/TYPO3-Documentation/t3SphinxThemeRtdDemoDocs/Textroles/Index.html

PR #92, opened 2016-09-15
Update 3-Command-controllers.rst

geraldspreer:
Some grammar and text changes


marble:
Thank you. Send more! Since I still saw a little error I corrected it. And once I was at it I tried to do a bit of more polishing.

PR #95, opened 2016-10-12
Delete one " (too many)

JavaJeremy:
Not sure, but I think when there are 3 of these something must be wrong, I'm just assuming this one is the wrong one!?


lolli42:
yep. looks good now. thx!

PR #96, opened 2016-10-15
QueryInterface:ORDER_ASCENDING needs two ":"

JavaJeremy:
I corrected the three entries I found that only had one ":"


lolli42:
thank yout!

PR #102, opened 2016-12-14
Persistence class mapping without leading \ in namespace

mueller-sebastian:
Removing leading \ from namespace in persistence class mapping demonstration


marble:
Thanks. Send more!

PR #108, opened 2017-02-02
Fix syntax error in code example

uwap:


uwap:
There are a few more of these in the chapter. I don't have the time to go over all of these but I would encourage that someone should do this.

marble:
Ok, I found two and fixed them.

PR #109, opened 2017-02-13
Typo and reminder for code example

kludikovsky:
* at Classes and Objects 
  * NULL instead of NUL
* at Constructor 
  * added Code Reminder


marble:
Thank you. Send more!

PR #110, opened 2017-03-08
[Doc] Remove wrong prefixed backslashes

hirnsturm:
The prefixed backslashes in namespace are wrong.


marble:
Thank you, "hirnsturm" :-)) Send more!

PR #111, opened 2017-03-09
Some typos and reST formulations chapter 3 - 5

mseremet:


marble:
Thanks a lot. 

(Here I use the single quote ' where I mean the "backtick":) Note that we don't have the textrole :class:'...' since this belongs to the Sphinx domain for Python. Use :php:'...' instead. Here's a description: https://docs.typo3.org/Tips/TipOfTheDay/Index.html#how-to-write-inline-code

Tip: Write '...' wherever you can instead of ' ' ... ' '.

I have not corrected those ':class:' occurrences yet. Feel free to to so.

mseremet:
Got it, Will keep that in mind for the next commits.

I was using other pages as template and the typographic conventions of this book, that's why I used them.

PR #112, opened 2017-03-22
Removes leading backslash in namespace declaration

funkvn:
This was wrong.


marble:
Very good, thanks. Send more!

PR #121, opened 2017-05-26
Fix makeInstance

CDRO:
makeInstance has a leading backslash, which is not allowed anymore.

Also t3lib_div does not exist anymore.


marble:
Thank you. Send more!

PR #128, opened 2017-08-29
Update 5-controlling-the-flow.rst

r12r:
Fix a little mistake. The InventoryController class contains the method listAction() instead of findAll().


marble:
Thank you. Send more!

PR #131, opened 2017-10-23
Replace confusing mentions of SVN with git

adimit:


marble:
Thank you. Send more!

adimit:
Thanks for the fast response.

You'll get a lot more, mainly on grammar. I'm reading through the documentation right now, making changes as I go. :-)

PR #136, opened 2017-10-26
Replace old @dontvalidate with @ignorevalidation

Lagerregal:
Replace old @dontvalidate with @ignorevalidation - 
@dontvalidate was syntax of the old propertymapper.


marble:
Thank you. Send more!

PR #137, opened 2017-11-09
Corrected a word.

Karthons:
"we don't have to thought about..." =>  "we don't have to think about..."


marble:
Thanks. Send more!

PR #149, opened 2018-01-25
Fix basic include of Includes.txt

carstendietrich:
Includes.txt was not correctly included, remove waste 'g'


marble:
Thank you. Send more!

PR #159, opened 2018-03-08
removed unnecessary call_user_func

wolffc:
by calling configurePlugin directly the example gets clearer


marble:
I didn't investigate - but for me it seems the code is on purpose there, creating a 'closure'. Can you check that? A closure creates a separate namespace.

DanielSiepmann:
You're right @marble, still in this example there is no local variable and therefore no need for a closure.

Issue #160, opened 2018-03-09
Extension configuration files


vojtasvoboda:
At this page: https://docs.typo3.org/typo3cms/ExtbaseFluidBook/4-FirstExtension/2-create-folder-structure-and-configuration-files.html

There is mentioned: "So that the extension can be loaded by TYPO3, we require two configuration files.".

But after that, there is only `ext_emconf.php` file. Which is the second configuration file?

PR #161, opened 2018-03-13
BUGFIX: Fix documentation about necessary configuration files

DanielSiepmann:
There is only a single necessary configuration file per extension.
Only ext_emconf.php is necessary and already documented.

Accordingly, references about a second file are removed.

Also paragraph was reformatted for line length.

Resolves: #160


marble:
Please review my proposals. I didn't understand the ((...)) thing with the icons. What should we write there?

PR #162, opened 2018-03-21
remove a misplaced quote

webfacer:
a misplaced quote which causes wrong rendering of the code-color


marble:
Thank you. Send more!

PR #166, opened 2018-05-01
Add tip to re-installation of extension

tantebootsy:
Especially for beginners it is important to mention that the extension has to be re-installed so that the controller is available on the CLI.

### DIFF
--- a/Documentation/10-Outlook/3-Command-controllers.rst
+++ b/Documentation/10-Outlook/3-Command-controllers.rst
@@ -56,8 +56,9 @@ After creation of the controller you need to register it. Add the following line
 Calling commands
 ----------------
 
-Once a command is registered and available, you can check the availability by calling
-`typo3/cli_dispatch.phpsh extbase help`.
+Re-install your extension to complete the registration. Once a command is registered and available, you can check the availability by calling
+`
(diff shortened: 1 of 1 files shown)


DanielSiepmann:
As far as I know it's not necessary to re-install the extension.
All you have to do is to clear the cache for `ext_localconf.php`, e.g. system cache.

Might you adjust the change? Otherwise I'll provide the necessary information.

Thanks in any case for mentioning that this hint is missing in the docs and providing an update.

tantebootsy:
Ah ok. Although I'm admin I sometimes don't see the system-cache-button. As the wordings and buttons for cache-clearing seem to change every major TYPO3-version maybe we should simply stick to  "clear backend cache" to at least distinguish between fe-cache and be-cache. Ok?

DanielSiepmann:
Ok.

The "cache" for `ext_localconf.php` is located at `typo3temp/**/ext_localconf_*.php`. The concrete path depends on the TYPO3 Version.


PS: You have to enable the system clear cache, even if you are admin, this is a global setting available in the install tool.

DanielSiepmann:
Thanks for your work :)

tantebootsy:
Thanks for YOUR work :)

PR #167, opened 2018-05-04
Removed quotation marks from default integers

karlsve:
Removed the quotation marks from the default values for integers, since this could lead to issues in newer MySQL versions.


marble:
Thank you. Send more!

PR #172, opened 2018-05-17
Update Index.rst

pi-phi:
missing dot after vendor name for configureplugin()


DanielSiepmann:
Good catch, thank you @pi-phi

PR #174, opened 2018-05-28
TASK: Fix $_EXTKEY usage in docs

DanielSiepmann:
As $_EXTKEY is more or less deprecated and will be removed in the
future. It's recommended to use the extension key as a string.

Therefore all examples were adjusted, including all mentions of the
variable.

### DIFF
--- a/Documentation/10-Outlook/2-Backend-modules.rst
+++ b/Documentation/10-Outlook/2-Backend-modules.rst
@@ -15,7 +15,7 @@ root directory of our extension. Lets look at our blog example how it defines a 
 
     if (TYPO3_MODE === 'BE') {
         \TYPO3\CMS\Extbase\Utility\ExtensionUtility::registerModule(
-            'Vendor.ext_key',
+            'Vendor.ExtensionName',
             'web',          // Main area
             'mod1',         // Name of the module
             '',             // Position of the module
--- a/Documentation/10-Outlook/2-Backend-modules.rst
+++ b/Documentation/10-Outlook/2-Backend-modules.rst
@@ -15,7 +15,7 @@ root directory of our extension. Lets look at our blog example how it defines a 
 
     if (TYPO3_MODE === 'BE') {
         \TYPO3\CMS\Extbase\Utility\ExtensionUtility::registerModule(
-            'Vendor' . $_EXTKEY,
+            'Vendor.ext_key',
             'web',          // Main area
             'mod1',         // Name of the module
             '',             // Position of the module
@@ -27,7 +27,7 @@ root directory of our extension. Lets look at our blog example how it defines a 
             array(          // Additional configuration
                 'access'    => 'user,group',
                 'icon'      => 'EXT:blog_example/ext_icon.gif',
-                'labels'    => 'LLL:EXT:' . $_EXTKEY . '/Resources/Private/Language/locallang_mod.xml',
+                'labels'    => 'LLL:EXT:ext_key/Resources/Private/Language/locallang_mod.xml',
             )
         );
     }
--- a/Documentation/10-Outlook/3-Command-controllers.rst
+++ b/Documentation/10-Outlook/3-Command-controllers.rst
@@ -47,7 +47,7 @@ After creation of the controller you need to register it. Add the following line
     <?php
 
     if (TYPO3_MODE === 'BE') {
-        $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['extbase']['commandControllers']['ext_key'] =
+        $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['extbase']['commandControllers']['ExtensionName-MeaningFullName'] =
             \Vendor\ExtKey\Command\SimpleCommandController::class;
     }
 
--- a/Documentation/10-Outlook/3-Command-controllers.rst
+++ b/Documentation/10-Outlook/3-Command-controllers.rst
@@ -47,8 +47,8 @@ After creation of the controller you need to register it. Add the following line
     <?php
 
     if (TYPO3_MODE === 'BE') {
-        $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['extbase']['commandControllers'][$_EXTKEY] =
-            \Vendor\Example\Command\SimpleCommandController::class;
+        $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['extbase']['commandControllers']['ext_key'] =
+            \Vendor\ExtKey\Command\SimpleCommandController::class;
     }
 
 .. _extbase_command_controller_calling:
--- a/Documentation/4-FirstExtension/2-create-folder-structure-and-configuration-files.rst
+++ b/Documentation/4-FirstExtension/2-create-folder-structure-and-configuration-files.rst
@@ -69,7 +69,7 @@ extension. Find a complete reference in chapter
 
 	<?php
 
-	$EM_CONF[$_EXTKEY] = [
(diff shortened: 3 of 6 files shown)


Documentation/10-Outlook/2-Backend-modules.rst
```diff
@@ -15,7 +15,7 @@ root directory of our extension. Lets look at our blog example how it defines a
 
     if (TYPO3_MODE === 'BE') {
         \TYPO3\CMS\Extbase\Utility\ExtensionUtility::registerModule(
-            'Vendor' . $_EXTKEY,
+            'Vendor.ext_key',
```
mbrodala: Please use `ExtensionName` instead to stay in line with the Extbase naming structure of plugins and modules.

Documentation/10-Outlook/3-Command-controllers.rst
```diff
@@ -47,8 +47,8 @@ After creation of the controller you need to register it. Add the following line
     <?php
 
     if (TYPO3_MODE === 'BE') {
-        $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['extbase']['commandControllers'][$_EXTKEY] =
-            \Vendor\Example\Command\SimpleCommandController::class;
+        $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['extbase']['commandControllers']['ext_key'] =
```
mbrodala: Here I'd recommend using a UNIX timestamp instead which allows for adding multiple command controllers for a single extension. Alternatively we could skip the key completely but having a key allows for easier unset in other locations.

Documentation/4-FirstExtension/2-create-folder-structure-and-configuration-files.rst
```diff
@@ -69,7 +69,7 @@ extension. Find a complete reference in chapter
 
 	<?php
 
-	$EM_CONF[$_EXTKEY] = [
+	$EM_CONF['ext_key'] = [
```
mbrodala: This is the only location where you should still use `$_EXTKEY` to be safe. There are older versions of the TYPO3 Composer Installers which only work with this variable.

Documentation/4-FirstExtension/7-configuring-the-plugin.rst
```diff
@@ -28,13 +28,15 @@ create in the top level of our extension directory.
...
 
-With the first line we prevent of security reasons, that the PHP code can be called directly outside of TYPO3.
-The static method ``configurePlugin()`` of the class offers several arguments.
-With the first we assign the extension key, which stand by in the global variable
-``$EXTKEY`` (it follows from the name of the extension directory) prefixed by the vendor namespace followed by a dot. This indicates, that we use namespaces for our php classes.
-With the second argument we give an unique name for the plugin (in UpperCamelCase notation).
-Because of historical reasons there is often used ``Pi1``, but maybe it is better to use more meaningfull
-names like "InventoryList".
+With the first line we prevent of security reasons, that the PHP code can be
+called directly outside of TYPO3.  The static method ``configurePlugin()`` of
```
mbrodala: Please add the class name to the method name. Also I think it should be this instead:

    :php:`ExtensionUtility::configurePlugin()`

This way the syntax language is absolutely clear and won't change on global changes.

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -50,7 +50,7 @@ backend. Let's have a look at the following two files:
 
     $pluginName = 'ExamplePlugin';
     \TYPO3\CMS\Extbase\Utility\ExtensionUtility::configurePlugin(
-        'Vendor.' . $_EXTKEY,
+        'Vendor.ext_key',
```
mbrodala: See above.

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -98,7 +95,7 @@ frontend plugin within the files :file:`ext_localconf.php` and :file:`Configurat
 
     $pluginName = 'Blog';
     \TYPO3\CMS\Extbase\Utility\ExtensionUtility::configurePlugin(
-        'Vendor.' . $_EXTKEY,
+        'Vendor.ext_key',
```
mbrodala: Here too.

Documentation/4-FirstExtension/2-create-folder-structure-and-configuration-files.rst
mbrodala: Not exactly but as said, only with [older TYPO3 Composer Installers](https://github.com/TYPO3/CmsComposerInstallers/pull/59).

Documentation/10-Outlook/3-Command-controllers.rst
DanielSiepmann: I'll change that, thanks. But I'm totally against unix timestamps, they are not helpfull, instead I would use something like 'ExtName-helpfullName`, to make listing in Backend Module easier to read.

Documentation/4-FirstExtension/2-create-folder-structure-and-configuration-files.rst
DanielSiepmann: Just read the link @sypets put into slack where this info was given, thanks @mbrodala I'll change that back.

sypets: Big thanks for clearing this up @mbrodala  and @DanielSiepmann

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -80,14 +80,11 @@ same format as above, containing all the non-cached-actions.
...
 
-The extension key (`$extensionKey` or `$_EXTKEY`) and `$pluginName` must be completely identical to the definition
-in :file:`ext_localconf.php`. `$extensionKey` must be filled with the extension key by yourself, where `$_EXTKEY` is not available. `$backendTitle` defines the displayed name of the plugin in
-the Backend.
-
-.. note::
-
-   **Attention:** The variable :php:`$_EXTKEY` is not available in the directory :file:`Configuration/TCA/`,
-   so you have to fill in your extension key directly.
+The extension key (`$extensionKey` or `example_extensionkey`) and `$pluginName`
```
mbrodala: I guess the `$extensionKey` variable explanation can be dropped here since it's not in the examples at all.

Personally I'd drop all variables here and use explanatory inline strings instead but that's out of scope of this change.

DanielSiepmann: Thanks, I've updated that section.

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -113,18 +106,18 @@ frontend plugin within the files :file:`ext_localconf.php` and :file:`Configurat
...
         'EXT:blog/Resources/Public/Icons/Extension.svg'
     );
 
-The plugin name (`$pluginName`) is ``Blog``. It is important that the name is exactly the same in
-:file:`ext_localconf.php` and :file:`Configuration/TCA/Overrides/tt_content.php`. The default action is ``index`` of controller
-``blog`` since it's the first element defined in the array and the first action in the list.
+The plugin name is ``Blog``. It is important that the name is exactly the same
+in :file:`ext_localconf.php` and
+:file:`Configuration/TCA/Overrides/tt_content.php`. The default action is
+``index`` of controller ``blog`` since it's the first element defined in the
```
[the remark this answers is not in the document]
DanielSiepmann: I've updated both paragraphs to use the php information, like method and FQCN instead.

Issue #173, opened 2018-05-17
In section: Developing a custom viewhelper add information about $escapeOutput


sypets:
Between TYPO3 7 and 8 the default behaviour of ViewHelpers was changed:

before: HTML code was output as is
after: HTML output is escaped: `<` becomes `&lt;`

The default behaviour can be changed:

```
/**
     * By default the output is escaped.
     *
     * @var bool
     */
    protected $escapeOutput = false;
```
In Fluid  f:format.htmlentities, f:format.stripTags etc. can be applied.

see : https://typo3.slack.com/archives/C025BQLFA/p1499087019054498

DanielSiepmann:
@lolli42 or @mbrodala might you correct or confirm my last comment? I'll update the docs accordingly afterwards.

mbrodala:
Yes, escaping always existed but before TYPO3v8 there was a single property `$escapingInterceptorEnabled` which was used to control the whole escaping behavior.

Since TYPO3v8 and standalone Fluid we have `$escapeOutput` and `$escapeChildren` instead.

PR #180, opened 2018-06-15
Feature: Improve custom ViewHelper

DanielSiepmann:
Overwork whole text.
Adjust Code Examples accordingly to PSR-2.
Add missing section about escaping.

Resolves: #173

### DIFF
--- a/Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
+++ b/Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
@@ -6,7 +6,7 @@ Developing a custom ViewHelper
 ==============================
 
 The development of an own ViewHelper is much asked for in practice and is part
-of the base repertoire of the extension development. This chapter will provide
+of the base repertoire of extension development. This chapter will provide
 step by step guidance for a small example from the blog example and describe
 enhanced techniques afterwards.
 
@@ -15,7 +15,7 @@ enhanced techniques afterwards.
 The Gravatar-ViewHelper
 -----------------------
 
-Avatar-Images are pictures or icons that for example are dedicated to the author
+"Avatar" images are pictures or icons that for example are dedicated to the author
 of an article in blogs or on forums. The photos of blog authors and forum
 moderators are mostly stored on the appropriate server. With users that only
 want to ask a question or to comment a blog post, this is not the case. To allow
@@ -25,7 +25,7 @@ certain avatar picture.
 
 A web application that wants to check if an avatar picture exists for a given
 email address has to send a checksum (with the hash function :php:`md5`) of the
-email address to the service and receives the picture to display.
+email address to the service and receives the picture for display.
 
 This section explains how to write a ViewHelper that uses an email address as
 parameter and shows the picture from gravatar.com if it exists.
@@ -36,19 +36,19 @@ Preliminary considerations
 --------------------------
 
 The first step should be thinking about how to use the ViewHelper later on in
-the template, in order to get a clear view about the arguments of the
+the template, in order to get a clear vision about the arguments of the
 ViewHelper.
 
 The ViewHelper is not part of the default distribution, therefore an own
 namespace import is necessary to use the ViewHelper. In the following example
 the namespace :php:`MyVendor\BlogExample\ViewHelpers` is imported with the
-prefix ``blog``. Now, all tags starting with ``blog:`` are interpreted as
+prefix `blog`. Now, all tags starting with `blog:` are interpreted as
 ViewHelper from within the namespace::
 
    {namespace blog=MyVendor\BlogExample\ViewHelpers}
 
-The ViewHelper should get the name gravatar and only get an email address as
-parameter. The ViewHelper is called in the template as follows:
+The ViewHelper should be given the name "gravatar" and only take an email
+address as parameter. The ViewHelper is called in the template as follows:
 
 .. code-block:: html
 
@@ -64,15 +64,14 @@ and the name of the XML element. The classname consists of the following three
 parts:
 
 * full namespace (in our example :php:`MyVendor\BlogExample\ViewHelpers`)
-* the name of the ViewHelper in UpperCamelCase writing (in our example ``Gravatar``)
-* the ending ``ViewHelper``
+* the name of the ViewHelper in UpperCamelCase writing (in our example `Gravatar`)
+* the ending `ViewHelper`
 
 For the Gravatar ViewHelper the name of the class is
 :php:`MyVendor\BlogExample\ViewHelpers\GravatarViewHelper`.
 
-Following the naming conventions for Extbase extensions the ViewHelper
-skeleton is created in the PHP file
-:file:`EXT:blog_example/Classes/ViewHelpers/GravatarViewHelper.php`::
+Following the naming conventions for Extbase extensions the ViewHelper skeleton
+is created in the PHP file :file:`EXT:blog_example/Classes/ViewHelpers/GravatarViewHelper.php`::
 
    <?php
    namespace MyVendor\BlogExample\ViewHelpers;
@@ -95,8 +94,9 @@ Every ViewHelper must inherit from the class
    A ViewHelper can also inherit from subclasses of :php:`AbstractViewHelper`,
    e.g.  from :php:`TYPO3\CMS\Fluid\Core\ViewHelper\AbstractTagBasedViewHelper`.
    Several subclasses are offering additional functionality. The
-   :php:`TagBasedViewHelper` will be explained later on in this chapter in
-   detail in ":ref:`creating-xml-tags-using-tagbasedviewhelper`".
+   :php:`TagBasedViewHelper` will be explained
+   :ref:`later on in this chapter <creating-xml-tags-using-tagbasedviewhelper>`
+   in detail.
 
 In addition every ViewHelper needs a method :php:`render()`, which is called
 once the ViewHelper is to be displayed in the template. The return value of the
@@ -116,7 +116,7 @@ extended like the following::
       }
    }
 
-And the ViewHelper is called in template like:
+This is how the ViewHelper is called used in the template:
 
 .. code-block:: html
 
@@ -124,21 +124,21 @@ And the ViewHelper is called in template like:
 
    Hello <blog:gravatar />
 
-Then ``Hello World`` is displayed.
+As result `Hello World` is displayed.
 
 .. _register-arguments-of-viewhelpers:
 
 Register arguments of ViewHelpers
 ---------------------------------
 
-The ``Gravatar`` ViewHelper must hand over the email address it should work on.
+The `Gravatar` ViewHelper must hand over the email address it should work on.
 This is the last needed building block, before the implementation of
 functionality can happen.
 
-All arguments of a ViewHelper must be registered. Every ViewHelper has to declare
-explicit which parameters are accepted.
+All arguments of a ViewHelper must be registered. Every ViewHelper has to
+declare explicitly which parameters are accepted.
 
-The one way to register these arguments is to enhance the :php:`render()`
+One way to register these arguments is to enhance the :php:`render()`
 method. All method arguments of the :php:`render()` method are automatically
 arguments of the ViewHelper. In the example it looks like this::
 
@@ -149,25 +149,24 @@ arguments of the ViewHelper. In the example it looks like this::
    {
    }
 
-This way the ViewHelper receives the argument ``emailAddress``, which is of the
-type ``string``. The type is fetched from the annotation of the method in the
-PHPDoc block.
+This way the ViewHelper receives the argument `emailAddress` of type `string`.
+The type is given by the annotation of the method in the PHPDoc block.
 
 .. warning::
 
    If the type of a parameter is not defined, an error message will be
-   displayed. The PHPDoc block has to be complete and syntactical correct.
-   For example, if ``@`` is missing in front of the ``param``, the type of the
-   parameter is not identified.
+   displayed. The PHPDoc block has to be complete and syntactical correct. For
+   example, if `@` is missing in front of the `param`, the type of the
+   parameter cannot be determined.
 
 .. tip::
 
-   Sometimes arguments should get different types. In this case the type mixed
+   Sometimes arguments can take various types. In this case the type `mixed`
    should be used in the PHPDoc. With the line :php:`@param mixed
-   $emailAddress` any type of object can be given as parameter ``emailAddress``,
+   $emailAddress` any type of object can be passed as parameter `emailAddress`,
    e.g. arrays, strings or integer values.
 
-At the end the output of :html:`img`-Tag is implemented::
+Finally the output of :html:`img`-Tag needs to be implemented::
 
    <?php
    namespace MyVendor\BlogExample\ViewHelpers;
@@ -195,12 +194,12 @@ Register Arguments with initializeArguments()
 ---------------------------------------------
 
 Initializing the ViewHelper arguments directly at the :php:`render()` method is
-extreme handy, when there aren't much arguments. But sometimes a complex
+extremely handy when there aren't many arguments. But sometimes a complex
 inheritance hierarchy is implemented with the ViewHelper, where different level
 of the inheritance structure should register additional arguments. Fluid itself
-does this for example with the ``form`` ViewHelpers.
+does this for example with the `form` ViewHelpers.
 
-Because method parameter and annotations are not inheritable, there must be an
+Because method parameters and annotations are not inheritable, there must be an
 additional way to register the arguments of a ViewHelper. Fluid provides the
 method :php:`initializeArguments()` for this. In this method additional
 arguments can be registered by calling :php:`$this->registerArgument($name,
@@ -231,7 +230,7 @@ identical::
       }
    }
 
-In this example the usage of :php:`initializeArguments()` is not particular
+In this example the usage of :php:`initializeArguments()` is not particularly
 meaningful, because the method only requires one parameter. When working with
 complex ViewHelpers which have a multilevel inheritance hierarchy, it is
 sometimes more readable to register the arguments with
@@ -248,11 +247,10 @@ ViewHelper provides a *Tag-Builder* that can be used to create tags. It takes
 care about the syntactical correct creation of the tag and escapes for example
 single and double quote in attributes.
 
-.. tip::
+.. attention::
 
-   With the correct escaping of the attributes the system security is enhanced,
-   because it prevents cross site scripting attacks that would break out of the
-   attributes of XML tags.
+   Correctly escaping the attributes is mandatory as it affects security and
+   prevents cross site scripting attacks.
 
 In the next step the example :php:`GravatarViewHelper` is modified a bit to use
 the :php:`TagBasedViewHelper`. Because the Gravatar-ViewHelper creates an
@@ -260,32 +258,34 @@ the :php:`TagBasedViewHelper`. Because the Gravatar-ViewHelper creates an
 
 .. todo:: Add code
 
-What has changed? First of all, the ViewHelper inherits not directly from
-:php:`AbstractViewHelper` but from :php:`TagBasedViewHelper`, which provides and
-initializes the Tag-Builder. Beyond that there is a class property
+What is different in this code?
+
+First of all, the ViewHelper does not inherit directly from
+:php:`AbstractViewHelper` but from :php:`TagBasedViewHelper`, which provides
+and initializes the Tag-Builder. Beyond that there is a class property
 :php:`$tagName` which stores the name of the tag to be created. Furthermore the
 Tag-Builder is available at :php:`$this->tag`. It offers the method
 :php:`addAttribute()` to add new tag attributes. In our example the attribute
 `src` is added to the tag, with the value assigned one line above it. Finally
-the Tag-Builder offers a method :php:`render()` which generates and returns the tag
-which than is given back.
+the Tag-Builder offers a method :php:`render()` which generates and returns the
+tag which than is given back.
 
 .. tip::
 
-   It might be unclear why this code is better even though it is much longer. It
-   communicates the meaning much better and therefore it is preferred to the
-   first example, where the gravatar URL and the creating of the :html:`img`-Tag
-   was mixed.
+   Why is this code better even though it is much longer? It communicates the
+   meaning in a much better way, and therefore it is preferred to the first
+   example, where the gravatar URL and the creation of the :html:`img`-tag
+   were mixed.
 
 The base class :php:`TagBasedViewHelper` allows implementing ViewHelpers which
-returns a XML tag easier and cleaner and help to concentrate at the essential.
+return XML tags easier and cleaner and help to concentrate at the essential.
 
 Furthermore the :php:`TagBasedViewHelper` offers assistance for ViewHelper
-arguments that should recur direct and unchanged as tag attributes. These could
-be registered in :php:`initializeArguments()` with the method
+arguments that should recur directly and unchanged as tag attributes. These
+could be registered in :php:`initializeArguments()` with the method
 :php:`$this->registerTagAttribute($name, $type, $description, $required =
-false)`.  If support for the :html:`<img>` attribute ``alt`` should be provided
-in the ViewHelper, this can be done by initializing this in
+false)`. If support for the :html:`<img>` attribute :html:`alt` should be
+provided in the ViewHelper, this can be done by initializing this in
 :php:`initializeArguments()` in the following way::
 
    public function initializeArguments()
@@ -297,7 +297,7 @@ For registering the universal attributes id, class, dir, style, lang, title,
 accesskey and tabindex there is a helper method
 :php:`registerUniversalTagAttributes()` available.
 
-If support for the universal attributes should be provided and the ``alt``
+If support for the universal attributes should be provided and the `alt`
 attribute in the Gravatar ViewHelper the following :php:`initializeArguments()`
 method will be necessary::
 
@@ -313,7 +313,7 @@ method will be necessary::
 Insert optional arguments
 -------------------------
 
-All ViewHelper arguments so fare registered were required. By setting a default
+All ViewHelper arguments so far registered were required. By setting a default
 value for an argument in the method signature, the argument is automatically
 optional. When registering the arguments through :php:`initializeArguments()`
 the according parameter has to be set to :php:`false`.
@@ -337,7 +337,7 @@ The :php:`render()` method can be improved like this::
       return $this->tag->render();
    }
 
-With this setting of a default value, the ``size`` attribute becomes optional.
+With this setting of a default value, the `size` attribute becomes optional.
 
 .. _prepare-viewhelper-for-inline-syntax:
 
@@ -345,22 +345,22 @@ Prepare ViewHelper for inline syntax
 ------------------------------------
 
 So far the Gravatar ViewHelper has focussed on the tag structure of the
-ViewHelper. The ViewHelper was only with the tag syntax (because it returns a
-tag as well):
+ViewHelper. The ViewHelper was written using tag syntax which seemed obvious
+because it returns a tag itself:
 
 .. code-block:: html
 
    <blog:gravatar emailAddress="{post.author.emailAddress}" />
 
-Alternatively this sample can be written in the inline notation:
+Alternatively this sample can be written using inline notation:
 
 .. code-block:: html
 
    {blog:gravatar(emailAddress: post.author.emailAddress)}
 
-With this, the tag concept of the ViewHelper is mostly gone. One should see the
-Gravatar ViewHelper as a kind of post processor for an email address and would
-allow the following syntax:
+In this form the tag structure of the ViewHelper is not visible. One should see
+the Gravatar ViewHelper as a kind of post processor for an email address and
+would allow the following syntax:
 
 .. code-block:: html
 
@@ -369,9 +369,9 @@ allow the following syntax:
 Here the email address has the focus and the Gravatar ViewHelper is a converting
 step based on the email address.
 
-The syntax ``{post.author.emailAddress -> blog:gravatar()}`` is an alternative
-writing for ``<blog:gravatar>{post.author.emailAddress}</blog:gravatar>``. To
-support this the email address comes either from the argument ``emailAddress``
+The syntax `{post.author.emailAddress -> blog:gravatar()}` is an alternative
+writing for `<blog:gravatar>{post.author.emailAddress}</blog:gravatar>`. To
+support this the email address comes either from the argument `emailAddress`
 or, if it is empty, the content of the tag should be interpreted as email
 address.
 
@@ -398,8 +398,8 @@ Lets have a look at the new code of the :php:`render()` method::
    }
 
 This code section has the following effect: First the ViewHelper attribute
-``emailAddress`` becomes optional. If no ``emailAddress`` attribute is given,
-content of the tag is interpreted as email address. The rest of the code in
+`emailAddress` is made optional. If no `emailAddress` attribute is given,
+the content of the tag is interpreted as email address. The rest of the code is
 unchanged.
 
 .. tip::
--- a/Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
+++ b/Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
@@ -1,96 +1,90 @@
 .. include:: ../Includes.txt
 
 Developing a custom ViewHelper
-================================================
+==============================
 
-The development of an own ViewHelper is much asked for in practice and
-is part of the base repertoire of the extension development. We will guide
-you step by step through a simple example from the blog example and describe
-enhanced techniques afterwards.
-
+The development of an own ViewHelper is much asked for in practice and is part
+of the base repertoire of the extension development. We will guide you step by
+step through a simple example from the blog example and describe enhanced
+techniques afterwards.
 
 The Gravatar-ViewHelper
--------------------------------------------------
+-----------------------
 
-Avatar-Images are pictures or icons that for example are dedicated
-to the author of an article in blogs or on forums. The photos of blog
-authors and forum moderators are mostly stored on the appropriate server.
-With users that only want to ask a question or to comment a blog post,
-this is not the case. To allow them to supply their article with an icon,
-a service called *gravatar.com* is available. This
-online service makes sure that an email address is assigned to a certain
-avatar picture.
+Avatar-Images are pictures or icons that for example are dedicated to the author
+of an article in blogs or on forums. The photos of blog authors and forum
+moderators are mostly stored on the appropriate server.  With users that only
+want to ask a question or to comment a blog post, this is not the case. To allow
+them to supply their article with an icon, a service called *gravatar.com* is
+available. This online service makes sure that an email address is assigned to a
+certain avatar picture.
 
-A web application that wants to check if an avatar picture exists
-for a given email address has to send a checksum (with the hash function
-*md5*) of the email address to the service and receives
-the picture to display. Therefore the use of
-*gravatar.com* introduces no security risk because the
-user of the blog only see the checksums of the email address and not the
-email address itself. This is possible as no efficient possibility is
-known to get the original data reconstructed from the checksum.
+A web application that wants to check if an avatar picture exists for a given
+email address has to send a checksum (with the hash function *md5*) of the email
+address to the service and receives the picture to display. Therefore the use of
+*gravatar.com* introduces no security risk because the user of the blog only see
+the checksums of the email address and not the email address itself. This is
+possible as no efficient possibility is known to get the original data
+reconstructed from the checksum.
 
-In this section we show you how to write your own ViewHelper that
-uses an email address as parameter and shows the picture from gravatar.com
-if it exists.
-
-
+In this section we show you how to write your own ViewHelper that uses an email
+address as parameter and shows the picture from gravatar.com if it exists.
 
 Preliminary considerations
--------------------------------------------------
+--------------------------
 
-The first step should be thinking about how to use the ViewHelper
-later on in the template, in order to get a clear view about the arguments
-of the ViewHelper. We take the point of view of a template author who
-wants to use our ViewHelper later on, without knowledge of the internal
-operations.
+The first step should be thinking about how to use the ViewHelper later on in
+the template, in order to get a clear view about the arguments of the
+ViewHelper. We take the point of view of a template author who wants to use our
+ViewHelper later on, without knowledge of the internal operations.
 
-First of all, think about how the ViewHelper should be called inside
-the template: The ViewHelper is not part of the default distribution,
-therefore we need an own namespace import to use the ViewHelper. We import
-the namespace ``MyVendor\BlogExample\ViewHelpers`` with the token
-``blog``. Now, all tags starting with ``blog:`` are
-interpreted as ViewHelper::
+First of all, think about how the ViewHelper should be called inside the
+template: The ViewHelper is not part of the default distribution, therefore we
+need an own namespace import to use the ViewHelper. We import the namespace
+``MyVendor\BlogExample\ViewHelpers`` with the token ``blog``. Now, all tags
+starting with ``blog:`` are interpreted as ViewHelper::
 
    {namespace blog=MyVendor\BlogExample\ViewHelpers}
 
-Our ViewHelper should get the name gravatar and only get an email
-address as parameter. We will call the ViewHelper in the template as
-follows::
+Our ViewHelper should get the name gravatar and only get an email address as
+parameter. We will call the ViewHelper in the template as follows:
+
+.. code-block:: html
 
    <blog:gravatar emailAddress="sebastian@typo3.org" />
 
 After this preliminary considerations we will start with the
 implementation.
 
-
-
 Now implementing!
--------------------------------------------------
+-----------------
 
-Every ViewHelper is a PHP class whose name is derived from the
-namespace import and the name of the XML element. The classname consists
-of the following three parts:
+Every ViewHelper is a PHP class whose name is derived from the namespace import
+and the name of the XML element. The classname consists of the following three
+parts:
 
-* full namespace (in our example
-  ``\MyVendor\BlogExample\ViewHelpers``)
-* the name of the ViewHelper in UpperCamelCase writing (in our
-  example ``Gravatar``)
+* full namespace (in our example ``\MyVendor\BlogExample\ViewHelpers``)
+* the name of the ViewHelper in UpperCamelCase writing (in our example ``Gravatar``)
 * the ending ``ViewHelper``
 
 For the Gravatar ViewHelper the name of the class is
 ``\MyVendor\BlogExample\ViewHelpers\GravatarViewHelper``.
 
-Following the naming conventions for Extbase extensions we create
-the ViewHelper skeleton in the PHP file
+Following the naming conventions for Extbase extensions we create the ViewHelper
+skeleton in the PHP file
 *EXT:blog_example/Classes/ViewHelpers/GravatarViewHelper.php*::
 
    <?php
    namespace MyVendor\BlogExample\ViewHelpers;
 
-   class GravatarViewHelper extends \TYPO3\CMS\Fluid\Core\ViewHelper\AbstractViewHelper {
-   public function render() {
-   }
+   use TYPO3\CMS\Fluid\Core\ViewHelper\AbstractViewHelper;
+
+   class GravatarViewHelper extends AbstractViewHelper
+   {
+      public function render()
+      {
+         // Implementation ...
+      }
    }
 
 Every ViewHelper must inherit from the class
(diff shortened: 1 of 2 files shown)


DanielSiepmann:
@mbrodala might you check this changes?

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,422 @@
...
+.. _developing-a-custom-viewhelper:
+
 Developing a custom ViewHelper
-================================================
+==============================
 
-The development of an own ViewHelper is much asked for in practice and
-is part of the base repertoire of the extension development. We will guide
-you step by step through a simple example from the blog example and describe
+The development of an own ViewHelper is much asked for in practice and is part
```
mbrodala: I'd say _frequently_ instead of _much_. Otherwise it sounds like _too much to ask for_. ;-)

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,422 @@
...
+want to ask a question or to comment a blog post, this is not the case. To allow
+them to supply their article with an icon, a service called *gravatar.com* is
+available. This online service makes sure that an email address is assigned to a
+certain avatar picture.
 
-In this section we show you how to write your own ViewHelper that
-uses an email address as parameter and shows the picture from gravatar.com
-if it exists.
+A web application that wants to check if an avatar picture exists for a given
+email address has to send a checksum (with the hash function :php:`md5`) of the
```
mbrodala: Please use `md5()` here.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,427 @@
...
-the template: The ViewHelper is not part of the default distribution,
-therefore we need an own namespace import to use the ViewHelper. We import
-the namespace ``MyVendor\BlogExample\ViewHelpers`` with the token
-``blog``. Now, all tags starting with ``blog:`` are
-interpreted as ViewHelper::
+The ViewHelper is not part of the default distribution, therefore an own
+namespace import is necessary to use the ViewHelper. In the following example
+the namespace :php:`MyVendor\BlogExample\ViewHelpers` is imported with the
+prefix ``blog``. Now, all tags starting with ``blog:`` are interpreted as
+ViewHelper from within the namespace::
```
mbrodala: Should we suggest to use the newer container/data-attribute variant instead?

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,427 @@
...
+method. All method arguments of the :php:`render()` method are automatically
+arguments of the ViewHelper. In the example it looks like this::
 
    /**
-   * @param string $emailAddress
-   */
-   public function render($emailAddress) {
+    * @param string $emailAddress
+    */
+   public function render($emailAddress)
```
mbrodala: This won't work in TYPO3v8 and newer, please switch to `initializeArguments()` instead.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: done

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Adjusted that, might you review again?

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
 .. tip::
 
-   A ViewHelper can also inherit from subclasses of
-   ``AbstractViewHelper``, e.g. from
-   ``\TYPO3\CMS\Fluid\Core\ViewHelper\AbstractTagBasedViewHelper``. Several
-   subclasses are offering additional functionality. We will talk about the
-   just addressed TagBasedViewHelper later on in this chapter in detail in
-   "Creating XML tags using TagBasedViewHelper".
+   A ViewHelper can also inherit from subclasses of :php:`AbstractViewHelper`,
+   e.g.  from :php:`TYPO3\CMS\Fluid\Core\ViewHelper\AbstractTagBasedViewHelper`.
```
mbrodala: @Namelesscoder is this still valid or should the one from standalone Fluid be preferred?

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+
+.. code-block:: html
 
    {namespace blog=MyVendor\BlogExample\ViewHelpers}
-   Hello <blog:gravatar />
 
-``Hello World`` should be displayed.
+   Hello <blog:gravatar />
 
+As result `Hello World` is displayed.
```
mbrodala: I don't think so, unless you change the code above. ;-)

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-
-   /**
-   * @return string the HTML <img>-Tag of the gravatar
-   */
-   public function render() {
-   return '<img src="http://www.gravatar.com/avatar/' .
-   md5($this->arguments['emailAddress']) . '" />';
-   }
-   }
+Creating XML tags using TagBasedViewHelper
```
mbrodala: Should be HTML/XML since HTML is the target format most of the time.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+------------------------------------------
 
-In this example the usage of
-``initializeArguments`` is not particular meaningful, because the
-method only requires one parameter. When working with complex ViewHelpers
-which have a multilevel inheritance hierarchy, it is sometimes more
-readable to register the arguments with
-``initializeArguments()``.
+For ViewHelper that create XML tags Fluid provides an enhanced baseclass: the
```
mbrodala: Either HTML/XML or leave it off completely.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-Tag-Builder is available at ``$this->tag``. It offers the
-method ``addAttribute`` *(Attribute, Value)*
-to add new tag attributes. In our example we add the attribute
-``src`` to the tag, with the value assigned one line above it.
-Finally the Tag-Builder offers a method ``render()`` which
-generates and returns the tag which than is given back, because we want to
-insert it in the template.
+First of all, the ViewHelper does not inherit directly from
+:php:`AbstractViewHelper` but from :php:`TagBasedViewHelper`, which provides
+and initializes the Tag-Builder. Beyond that there is a class property
```
mbrodala: Should be:

    :php:`TagBuilder`

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+:php:`$tagName` which stores the name of the tag to be created. Furthermore the
+Tag-Builder is available at :php:`$this->tag`. It offers the method
```
mbrodala: If not referring to the class it should be _tag builder_.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+:php:`addAttribute()` to add new tag attributes. In our example the attribute
+`src` is added to the tag, with the value assigned one line above it. Finally
+the Tag-Builder offers a method :php:`render()` which generates and returns the
+tag which than is given back.
```
mbrodala: _... returns the tag markup._

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-
-   public function initializeArguments() {
-   $this->registerTagAttribute('alt', 'string', 'Alternative Text for the image');
+   Why is this code better even though it is much longer? It communicates the
+   meaning in a much better way, and therefore it is preferred to the first
+   example, where the gravatar URL and the creation of the :html:`img`-tag
+   were mixed.
+
+The base class :php:`TagBasedViewHelper` allows implementing ViewHelpers which
+return XML tags easier and cleaner and help to concentrate at the essential.
```
mbrodala: _... focus on ..._

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+
+Furthermore the :php:`TagBasedViewHelper` offers assistance for ViewHelper
+arguments that should recur directly and unchanged as tag attributes. These
```
mbrodala: _... should be outputted directly ..._

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+could be registered in :php:`initializeArguments()` with the method
```
mbrodala: _must be ..._

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-is a helper method ``registerUniversalTagAttributes()``
-available.
+For registering the universal attributes id, class, dir, style, lang, title,
+accesskey and tabindex there is a helper method
+:php:`registerUniversalTagAttributes()` available.
 
-If we want to support the universal attributes and the
-``alt`` attribute in our ``Gravatar`` ViewHelper we need
-the following ``initializeArguments()`` method::
+If support for the universal attributes should be provided and the `alt`
```
mbrodala: _... in addition to the ..._

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-   $gravatarUri = 'http://www.gravatar.com/avatar/' . md5($emailAddress) . '?s=' . urlencode($size);
-   $this->tag->addAttribute('src', $gravatarUri);
-   return $this->tag->render();
+    * @return string the HTML <img>-Tag of the gravatar
+    */
+   public function render()
+   {
+      $gravatarUri = 'http://www.gravatar.com/avatar/' .
+         md5($this->arguments['emailAddress']) .
+         '?s=' . urlencode($this->arguments['size']);
```
mbrodala: Maybe mention here that clamping the given `size` argument is a challenge to the reader? ;-)

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
    }
 
-This code section has the following effect: First we have
-made the ViewHelper attribute ``emailAddress`` optional. If no
-``emailAddress`` attribuite is given, we interpret the content of
-the tag as email address. The rest of the code in unchanged.
+This code section has the following effect: First the ViewHelper attribute
+`emailAddress` is made optional. If no `emailAddress` attribute is given,
+the content of the tag is interpreted as email address. The rest of the code is
+unchanged.
```
mbrodala: Should we mention the traits in Fluid which vastly simplify this?

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
NamelessCoder: It depends: if you must have the legacy support for render method arguments then you must use the CMS versions. For every other case I strongly recommend just using the Fluid-native classes as they are completely capable of rendering also CMS-specific ViewHelpers and so so more efficiently.

The broader perspective is that VH classes will work without a TYPO3 context if they do not depend on that category of TYPO3 classes that require this.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+.. _developing-a-custom-viewhelper:
+
 Developing a custom ViewHelper
-================================================
+==============================
 
-The development of an own ViewHelper is much asked for in practice and
-is part of the base repertoire of the extension development. We will guide
-you step by step through a simple example from the blog example and describe
+The development of an own ViewHelper is frequently asked for in practice and is
```
NamelessCoder: "The development of an own ViewHelper is frequently asked for in practice" is a sub-optimal wording that is in part caused by sentence syntax differences in English and German. Proper English wording would be something along the lines of:

"Developing a custom ViewHelper is often necessary in extension development" to replace the entire sentence.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+part of the base repertoire of extension development. This chapter will provide
+step by step guidance for a small example from the blog example and describe
```
NamelessCoder: "...example from the blog example..." is a little muddy. I suggest:

"This chapter will demonstrate how to write a custom ViewHelper for the blog example extension"

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
 enhanced techniques afterwards.
 
+.. _the-gravatar-viewhelper:
 
 The Gravatar-ViewHelper
```
NamelessCoder: You may want to remove the hyphen here. Either "Gravatar ViewHelper" or "GravatarViewHelper". Hyphen concatenations are much, much less frequent in English compared to European languages.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+email address to the service and receives the picture for display.
 
+This section explains how to write a ViewHelper that uses an email address as
+parameter and shows the picture from gravatar.com if it exists.
 
+.. _preliminary-considerations:
 
 Preliminary considerations
--------------------------------------------------
+--------------------------
```
NamelessCoder: I suggest dropping this entire section. What it explains can be pretty much explained with a single line of commenting in the VH example class source that gets shown.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+the template, in order to get a clear vision about the arguments of the
+ViewHelper.
 
-First of all, think about how the ViewHelper should be called inside
-the template: The ViewHelper is not part of the default distribution,
-therefore we need an own namespace import to use the ViewHelper. We import
-the namespace ``MyVendor\BlogExample\ViewHelpers`` with the token
-``blog``. Now, all tags starting with ``blog:`` are
-interpreted as ViewHelper::
+The ViewHelper is not part of the default distribution, therefore an own
```
NamelessCoder: Do take care to avoid the wording "an own XYZ" - it is not a proper English construct. It also exists in my own language, Danish, so I can see why it is tempting to do ;)

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+the namespace :php:`MyVendor\BlogExample\ViewHelpers` is imported with the
+prefix `blog`. Now, all tags starting with `blog:` are interpreted as
+ViewHelper from within the namespace::
 
    {namespace blog=MyVendor\BlogExample\ViewHelpers}
 
-Our ViewHelper should get the name gravatar and only get an email
-address as parameter. We will call the ViewHelper in the template as
-follows::
+The ViewHelper should be given the name "gravatar" and only take an email
```
NamelessCoder: It doesn't help understanding to allude to "ViewHelper name" here. Stick to just giving the class name and explaining how an NS import is converted to class name.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
 
 Now implementing!
--------------------------------------------------
+-----------------
 
-Every ViewHelper is a PHP class whose name is derived from the
-namespace import and the name of the XML element. The classname consists
-of the following three parts:
+Every ViewHelper is a PHP class whose name is derived from the namespace import
+and the name of the XML element. The classname consists of the following three
```
NamelessCoder: "class name", not classname.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+parts:
 
-* full namespace (in our example
-  ``\MyVendor\BlogExample\ViewHelpers``)
-* the name of the ViewHelper in UpperCamelCase writing (in our
-  example ``Gravatar``)
-* the ending ``ViewHelper``
+* full namespace (in our example :php:`MyVendor\BlogExample\ViewHelpers`)
```
NamelessCoder: You may not want to explain this here. ViewHelper resolving rules are much less strict now than they used to be: what matters is the local class name w/o namespace and that VH must be in a namespace you import. Some parts, like the ViewHelper suffix, are technically no longer required.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
NamelessCoder: Double space character has snuck into this line after the `e.g.  `.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
    <?php
    namespace MyVendor\BlogExample\ViewHelpers;
 
-   class GravatarViewHelper extends \TYPO3\CMS\Fluid\Core\ViewHelper\AbstractViewHelper {
-      public function render() {
+   use TYPO3\CMS\Fluid\Core\ViewHelper\AbstractViewHelper;
+
+   class GravatarViewHelper extends AbstractViewHelper
+   {
+      public function render()
```
NamelessCoder: Given that the render method is at this time optional to use, it is not the ideal demonstration. At least, the example VH here should be more thoroughly explained concerning optimisations like using renderStatic() or even compile().

The render method is the worst of all the options.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-``render()`` method are automatically arguments of the
-ViewHelpers. In our example it looks like this::
-
-   /**
-   * @param string $emailAddress
-   */
-   public function render($emailAddress) {
+---------------------------------
+
+The `Gravatar` ViewHelper must hand over the email address it should work on.
```
NamelessCoder: Don't end sentences with prepositions. Correct wording could be "...the email address on which it should operate" but I'd suggest using "...the email address which identifies the Gravatar".

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+This is the last needed building block, before the implementation of
```
NamelessCoder: Suggest: "The is the last remaining piece before the implementation can be completed"

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+functionality can happen.
+
+All arguments of a ViewHelper must be registered. Every ViewHelper has to
+declare explicitly which parameters are accepted. The registration happens
```
NamelessCoder: This is no longer true. On the surface, yes, but there are actually two additional options for accepting undeclared arguments, such as ones that use a prefix.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+inside `initializeArguments()`::
+
+   public function initializeArguments()
+   {
+      $this->registerArgument(
```
NamelessCoder: This creates an optional argument. You may want to add the fourth `true` parameter.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-.. warning::
-   If you forget to specify the type of a parameter, an error message
-   will be displayed. Check at all times that the PHPDoc block is complete
-   and syntactical correct. For example, if you forget the ``@``
-   in front of the ``param``, the type of the parameter is not
-   identified.
+This way the ViewHelper receives the argument `emailAddress` of type `string`.
+The type is given by the second argument of the `registerTagAttribute()`.  The
+arguments are registered via :php:`$this->registerArgument($name, $type,
+$description, $required, $defaultValue)`. These arguments can be accessed
```
NamelessCoder: Accessing arguments depends on context. Static calling means you get `$arguments` and will cause PHP errors if you use `$this->arguments`.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-   Sometimes arguments should get *different*
-   types. In this case you should use the type mixed in the PHPDoc. With
-   the line ``@param mixed $emailAddress`` any type of object can
-   be given as parameter ``emailAddress``, e.g. arrays, strings or
-   integer values.
+   Sometimes arguments can take various types. In this case the type `mixed`
+   should be used.
 
-At the end we implement the output as img tag::
+Finally the output of :html:`img`-Tag needs to be implemented::
```
NamelessCoder: "Tag" should be written in lowercase. The wording should be "Finally, the output of the img tag needs to be implemented".

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
NamelessCoder: "For ViewHelpers which create HTML/XML tags, Fluid provides an enhanced base class"

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+------------------------------------------
 
-In this example the usage of
-``initializeArguments`` is not particular meaningful, because the
-method only requires one parameter. When working with complex ViewHelpers
-which have a multilevel inheritance hierarchy, it is sometimes more
-readable to register the arguments with
-``initializeArguments()``.
+For ViewHelper that create XML tags Fluid provides an enhanced baseclass: the
+:php:`TYPO3\CMS\Fluid\Core\ViewHelper\AbstractTagBasedViewHelper`. This
```
NamelessCoder: Wrong parent class unless you are intentionally demanded to support legacy render method arguments.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+ViewHelper provides a *Tag-Builder* that can be used to create tags. It takes
```
NamelessCoder: TagBuilder, or tag builder. Not Tag-Builder.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+care about the syntactical correct creation of the tag and escapes for example
```
NamelessCoder: "It takes care of the syntactically correct creation of tags and for example escapes single and double quotes in attribute values".

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+.. attention::
 
+   Correctly escaping the attributes is mandatory as it affects security and
+   prevents cross site scripting attacks.
 
-Creating XML tags using TagBasedViewHelper
---------------------------------------------------------------------------------------------------
+In the next step the example :php:`GravatarViewHelper` is modified a bit to use
+the :php:`TagBasedViewHelper`. Because the Gravatar-ViewHelper creates an
+:html:`img`-Tag the use of the Tag-Builder is advised.
```
NamelessCoder: Tag should not be uppercase, I personally would not concatenate this either. ":html:`img` tag" is viable. Do not concatenate Gravatar-ViewHelper; one word or two.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
NamelessCoder: Suggest: "Finally, the ``GravatarViewHelper`` creates an ``img`` tag builder has a method named :php:`render()` which generates and returns the tag after configuring the tag builder instance."

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-could be registerd in ``initializeArguments()`` with the method
-``$this->registerTagAttribute($name, $type, $description, $required
-= FALSE)``. If we want to support the ``<img>``
-attribure ``alt`` in our ViewHelper, we can initialize this in
-``initializeArguments()`` in the following way::
-
-   public function initializeArguments() {
-   $this->registerTagAttribute('alt', 'string', 'Alternative Text for the image');
+   Why is this code better even though it is much longer? It communicates the
+   meaning in a much better way, and therefore it is preferred to the first
```
NamelessCoder: "therefore it is preferable to the first example"

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+   example, where the gravatar URL and the creation of the :html:`img`-tag
+   were mixed.
```
NamelessCoder: Suggest: "were mixed together".

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
NamelessCoder: Suggest: remove the entire sentence tail so it ends with "which make creating XML tags easier".

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
NamelessCoder: "Should be output directly".  Though it is technically correct to use "outputted", a massive 90%+ of native English speakers will without thinking prefer to use "output" as past tense of... to output.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
NamelessCoder: "If support for universal attributes"

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-
-Back to our example: We can add a size parameter for the picture in
-the Gravatar ViewHelper. This size parameter will be used to determine the
-height and width of the image in pixels and can range from 1 to 512. When
-no size is given, an image of 80px is generated.
-
-We can enhance the ``render()`` method like this::
+-------------------------
+
+All ViewHelper arguments so far registered were required. By setting a default
```
NamelessCoder: "All ViewHelper arguments which were registered until now have been required".

Also, this statement is actually false given the code examples. They create optional arguments.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+value for an argument, the argument becomes optional.  When registering the
```
NamelessCoder: This is not how you make an argument optional. You flag it as such with the fourth parameter; default value or no default value makes zero difference (though logically it's void to use a default value for a mandatory argument).

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+arguments through :php:`initializeArguments()` the according parameter has to be
+set to :php:`false`.
```
NamelessCoder: Not true. Opposite: to make mandatory, pass true.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+
+Back to the example: An optional size parameter for the picture in the Gravatar
```
NamelessCoder: "An optional size for the image can be provided to the Gravatar ViewHelper." Be careful with sentence syntax: it doesn't work the same way as our native languages do!

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+ViewHelper can be provided. This size parameter will be used to determine the
+height and width of the image in pixels and can range from 1 to 512. When no
```
NamelessCoder: "...the height and width in pixels of the image..."

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+         '?s=' . urlencode($this->arguments['size']);
+      $this->tag->addAttribute('src', $gravatarUri);
+      return $this->tag->render();
    }
-   }
-
-With this setting of a default value we have made the
-``size`` attribute optional.
 
+With this setting of a default value, the `size` attribute becomes optional.
```
NamelessCoder: Incorrect.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
 
+.. _prepare-viewhelper-for-inline-syntax:
 
 Prepare ViewHelper for inline syntax
---------------------------------------------------------------------------------------------------
+------------------------------------
+
+So far the Gravatar ViewHelper has focussed on the tag structure of the
+ViewHelper. The ViewHelper was written using tag syntax which seemed obvious
```
NamelessCoder: "The call to render the ViewHelper was written with tax syntax"

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+because it returns a tag itself:
```
NamelessCoder: "...because it itself returns a tag:"

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+.. code-block:: html
 
-So far with our gravatar ViewHelper we have focussed on the tag
-structure of the ViewHelper. We have used the ViewHelper only with the tag
-syntax (because it returns a tag as well):
+   <blog:gravatar emailAddress="{post.author.emailAddress}" />
 
-``<blog:gravatar emailAddress="{post.author.emailAddress}"
-/>``
+Alternatively this sample can be written using inline notation:
```
NamelessCoder: s/sample/expression

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+.. code-block:: html
 
-``{blog:gravatar(emailAddress:
-post.author.emailAddress)}``
+   {blog:gravatar(emailAddress: post.author.emailAddress)}
 
-With this, the tag concept of the ViewHelper is mostly gone. One
-should see the gravatar ViewHelper as a kind of post processor for an
-email address and would allow the following syntax:
+In this form the tag structure of the ViewHelper is not visible. One should see
```
NamelessCoder: "Using this syntax, the tag structure of the ViewHelper is not evident." but consider removing the leading sentence instead. The paragraph works well without it and it doesn't add much that isn't explained in more detail later.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+   {post.author.emailAddress -> blog:gravatar()}
 
-We want to show you now what a ViewHelper has to do, to support this
-syntax. The syntax ``{post.author.emailAddress ->
-blog:gravatar()}`` is an alternative writing for
-``<blog:gravatar>{post.author.emailAddress}</blog:gravatar>``.
-To support this we have to use the email address either from the argument
-``emailAddress`` or, if it is empty, we should interpret the
-content of the tag as email address.
+Here the email address has the focus and the Gravatar ViewHelper is a converting
```
NamelessCoder: Definitely reword. I suggest:

"This syntax places focus on the variable that is passed to the ViewHelper as it comes first." It is not necessary and slightly misleading to explain that the ViewHelper "does conversion". It takes one variable and outputs something else: a tag with certain properties. This semantically isn't a conversion in particular when you also consider the extra supported arguments for size etc.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+step based on the email address.
 
-How did we get the content of a ViewHelper tag? For this a helper
-method ``renderChildren()`` is available in the
-``AbstractViewHelper``. This returns the evaluated object between
-the opening and closing tag.
+The syntax `{post.author.emailAddress -> blog:gravatar()}` is an alternative
+writing for `<blog:gravatar>{post.author.emailAddress}</blog:gravatar>`. To
```
NamelessCoder: "is an alternative notation/syntax..." or "an alternative way of writing XYZ".

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+         '?s=' . urlencode($this->arguments['size']);
+      $this->tag->addAttribute('src', $gravatarUri);
+      return $this->tag->render();
    }
 
-This code section has the following effect: First we have
-made the ViewHelper attribute ``emailAddress`` optional. If no
-``emailAddress`` attribuite is given, we interpret the content of
-the tag as email address. The rest of the code in unchanged.
+This code section has the following effect: First the ViewHelper attribute
```
NamelessCoder: Consider removing these many references that explain the same thing. It'll make it easier to handle if there's only one or two instances of incorrect information about how you make mandatory arguments ;)

This code does _not_ have the effect of making the argument optional. It has the effect of adding a secondary source for reading the argument value. The `registerArgument` call is what makes it optional.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+`emailAddress` is made optional. If no `emailAddress` attribute is given,
```
NamelessCoder: s/attribute/argument

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+the content of the tag is interpreted as email address. The rest of the code is
```
NamelessCoder: Suggest: "If no email address argument is provided, Fluid tries to read the value from tag contents or inline passing of child node value".

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+unchanged.
 
 .. tip::
 
-   This trick was specially used at the format ViewHelpers. Every
-   ViewHelper supports both writings there.
+   This trick was specially used at the format ViewHelpers. Every ViewHelper
```
NamelessCoder: "This trick was in particular used with formatting ViewHelpers. These ViewHelpers all support both tag mode and inline syntax."

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+   supports both writings there.
+
+.. _escaping-of-output:
+
+Escaping of Output
+------------------
+
+By default all output is escaped to prevent cross site scripting. In above
+example no :html:`<img>`-Tag will be displayed, instead escaped HTML will be
```
NamelessCoder: remove concatenation before `-Tag` and write `tag` in lowercase.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
-   ViewHelper supports both writings there.
+   This trick was specially used at the format ViewHelpers. Every ViewHelper
+   supports both writings there.
+
+.. _escaping-of-output:
+
+Escaping of Output
+------------------
+
+By default all output is escaped to prevent cross site scripting. In above
```
NamelessCoder: "In the example above" or "In the above example" with the former being preferred by native speakers.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+example no :html:`<img>`-Tag will be displayed, instead escaped HTML will be
+displayed.
+
+To allow the raw output, the escaping has to be disabled. This is done by
```
NamelessCoder: "To allow unescaped output, escaping has to be explicitly disabled"

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,422 +1,400 @@
...
+setting the class property :php:`$escapeOutput` to `false`::
+
+   protected $escapeOutput = false;
 
+In addition to the output of the ViewHelper itself, there is another property
+defining the escaping of child ViewHelpers if any::
```
NamelessCoder: Not just child ViewHelpers but in fact, any type of child node. Suggest adding:

"If escaping of children is disabled, no nodes that you pass with inline syntax or values you use as tag content will be escaped. Note that escapeOutput takes priority: if it is disabled, escaping of child nodes is also disabled unless explicitly enabled."

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: I've removed the double whitespace, which is ignored during rendering but still disturbs while reading the source.

I've also just exchanged the class name with the one of TYPO3Fluid. It sounds like it's the recommended way. We might add some information about how to support legacy bases later on. If that's still necessary.

In my opinion we should concentrate on finishing this rework first. Otherwise we will never finish. It's ok if the result is better then what was written before.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: In the example the ViewHelper returns `World`, so `Hello World` looks right to me. What am I missing?

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: As far as I can see, the tags are always valid XML Tags. As you can also generate XML and HTML5 is not valid XML most of the time.

Still I added the information that this is also valid HTML.

DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: I've added the FQCN with `php` role.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Replaced in three occurrences.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: I've rephrased with both suggestions to:

```
Finally, the GravatarViewHelper creates an img tag builder, which has a method named
:php:`render()`. After configuring the tag builder instance, the rendered tag
markup is returned.
```

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Changed both.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Added blank lines to group lines of code.

Assigned arguments to variables before for better reading.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: I didn't work with these traits yet, do you have an example? If we mention them, we should also provide an example.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Thanks. I changed to a whitespace. As this is more a tutorial, the user might not be familiar with the class name. So using two words might be more friendly to the reader at this point.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done, removed the whole section.

We do not need to bloat the documentation. One would also not bloat code if not necessary.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Section was removed.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Section was removed.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Section was removed.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Section was removed.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Fixed all occurrences of `-Tag` on the document.

Also added `the`.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Replaced all occurrences of the class with the one of TYPO3Fluid.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Replaces all occurrences with either FQCN or `tag builder`.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Adjusted all occurrences in the document.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: I've removed the whole sentence and added the information in the last sentence of this section.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: I've removed this sentence and added the information to the last sentence in this section.

This was still from the old example with `render` method containing the arguments.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Removed this and added the information to the last sentence in this section. The wrong information was fixed.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Removed and added the correct information in last sentence of the section.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Removed the sentence.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: I've used syntax again, as it was suggested in a sentence above already.
This way we keep the same wording.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: This was changed in the whole document.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Used the preferred one.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: Done. Justed rephrased to not contain `you` and added syntax highlighting to `escaoeOutput`.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
DanielSiepmann: I've removed the first sentence, mentioning the optional behaviour.

sypets:
Hi. Can we first focus on making it technically correct? 

Refining grammar and style later shouldn't be a problem and can be done in one batch. We've even got a native speaker who may be willing to take a look once it's done.

tmotyl:
ping

marble:
@DanielSiepmann What should we do with this pull request? Is somebody willing to care and jump in as "assignee"? If not I suggest to close it for now.

tmotyl:
I've pushed rebased version to the https://github.com/TYPO3-Documentation/TYPO3CMS-Book-ExtbaseFluid/pull/194 as I don't have rights to edit this pull request.

DanielSiepmann:
Thanks @tmotyl, so we proceed in #194

PR #190, opened 2018-07-28
Link the latest repo of blog example and add Italien version of TYPO3…

fawmi:
… extbase book


DanielSiepmann:
Thanks for the update. Just one question before I merge this, is there a reason why you removed the link to the German version of the book?

fawmi:
I have not deleted the Link to german version of this book, but i made the link text easy to understand.
I changed (de) Extbasebuch to Extbasebuch (German) which makes more sense and is easy to understand.
image

PR #194, opened 2018-09-03
Feature: Update section about custom ViewHelper

tmotyl:
Provide up to date information about writing custom ViewHelper.

fixes https://github.com/TYPO3-Documentation/TYPO3CMS-Book-ExtbaseFluid/pull/180

### DIFF
--- a/Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
+++ b/Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
@@ -6,7 +6,7 @@ Developing a custom ViewHelper
 ==============================
 
 The development of an own ViewHelper is much asked for in practice and is part
-of the base repertoire of the extension development. This chapter will provide
+of the base repertoire of extension development. This chapter will provide
 step by step guidance for a small example from the blog example and describe
 enhanced techniques afterwards.
 
@@ -15,7 +15,7 @@ enhanced techniques afterwards.
 The Gravatar-ViewHelper
 -----------------------
 
-Avatar-Images are pictures or icons that for example are dedicated to the author
+"Avatar" images are pictures or icons that for example are dedicated to the author
 of an article in blogs or on forums. The photos of blog authors and forum
 moderators are mostly stored on the appropriate server. With users that only
 want to ask a question or to comment a blog post, this is not the case. To allow
@@ -25,7 +25,7 @@ certain avatar picture.
 
 A web application that wants to check if an avatar picture exists for a given
 email address has to send a checksum (with the hash function :php:`md5`) of the
-email address to the service and receives the picture to display.
+email address to the service and receives the picture for display.
 
 This section explains how to write a ViewHelper that uses an email address as
 parameter and shows the picture from gravatar.com if it exists.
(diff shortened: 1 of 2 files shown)


DanielSiepmann:
@namelesscoder can you check the updated information, whether they cover everything necessary and are correct?

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,442 +1,702 @@
...
+          \Closure $renderChildrenClosure,
+          RenderingContextInterface $renderingContext
+      ) {
+          // Implementation ...
+      }
    }
 
 Every ViewHelper must inherit from the class
-``\TYPO3\CMS\Fluid\Core\ViewHelper\AbstractViewHelper``.
+:php:`\TYPO3\CMS\Fluid\Core\ViewHelper\AbstractViewHelper`.
```
NamelessCoder: This should refer to the class `\TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper` - since dropping render method argument support the CMS-specific abstract class is an alias of the Fluid class. Same for all other abstract VH classes. There are several occurrences below this one.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,442 +1,702 @@
...
-``initializeArguments()``.
+   <img src="http://www.gravatar.com/avatar/5f0efb20de5ecfedbe0bf5e7c12353fe" />
 
+Instead the result is:
 
+.. code-block:: html
 
-Creating XML tags using TagBasedViewHelper
---------------------------------------------------------------------------------------------------
+   &lt;img src=&quot;http://www.gravatar.com/avatar/5f0efb20de5ecfedbe0bf5e7c12353fe&quot; /&gt;
```
NamelessCoder: I think there's an entity encoding problem here?

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,442 +1,702 @@
...
+       {
+           $this->registerArgument('emailAddress', 'string', 'The email address to resolve the gravatar for', false, null);
+       }
+
+       public static function renderStatic(
+           array $arguments,
+           \Closure $renderChildrenClosure,
+           RenderingContextInterface $renderingContext
+       ) {
+           $emailAddress = $arguments['emailAddress'] ?? $renderChildrenClosure();
```
NamelessCoder: This tip should be rewritten to a combination of:

* Using `CompileWithContentArgumentAndRenderStatic` trait.
* Simply calling `$emailAddress = $renderChildrenClosure();` to read the value.
* Dropping the `, false, null` part in argument registrations since these are the default parameter values and are redundant to specify.

Since this "use argument or child content" is a Fluid-native feature we should demonstrate how to use it.

Note: in the future this is expected to be replaced with being able to declare `protected $contentArgumentName = 'emailAddress';` and the VH execution would take care of the rest. Unfortunately that's not yet possible: the first optional argument will be assumed to be the content argument.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
```diff
@@ -1,442 +1,702 @@
...
-``emailAddress`` attribute is given, we interpret the content of
-the tag as email address. The rest of the code in unchanged.
+.. _renderstatic-method:
 
-.. tip::
+:php:`renderStatic()`-Method
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Most of the time this method is implemented. It's the one that is called by
+default from within the compiled Fluid.
```
NamelessCoder: I would add this sentence after the first one: 

> It is however not called on AbstractTagBasedViewHelper implementations - on such classes you still need to use the `render()` method since that is the only way you can access `$this->tag` which contains the tag builder that generates the actual XML tag.

Documentation/8-Fluid/8-developing-a-custom-viewhelper.rst
lolli42: no. it's an escaping example

tmotyl:
Thanks a bunch @lolli, i didnt manage to finish up this one before going holidays.

PR #197, opened 2018-10-03
Update example code to be PSR2 compliant

benabbottnz:
This change will show the code with tabs of four spaces making it PSR2 compliant. Without this change you are presented with code that has random spacing consisting of two and three spaces, which isn't PSR2 compliant.

### DIFF
--- a/Documentation/4-FirstExtension/2-create-folder-structure-and-configuration-files.rst
+++ b/Documentation/4-FirstExtension/2-create-folder-structure-and-configuration-files.rst
@@ -70,25 +70,25 @@ extension. Find a complete reference in chapter
    <?php
 
    $EM_CONF[$_EXTKEY] = [
-      'title' => 'Store Inventory',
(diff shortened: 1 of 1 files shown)


marble:
Thank you very much. Send more! Note: In documentation we generally follow this strategy: One indentation level is three blanks. Rationale: ① documentation is not coding, it is documenting that include examples. ② Indentation should be enough to be well visible but not more than necessary. ③ Indenting by three nicely fits with many reST constructs. ④ We have code example for many different languages (xml, html, php, js, typescript, typoscript, ...) We don't want want those many different coding guidelines. ⑤ A simple rule like "one level == three spaces" allows easy automatic normalizing of the documentation sources.

That being said: We accept examples with indentation of 4 spaces.  But once somebody change that to three we won't change back to four.

benabbottnz:
Make sense! Thanks for the info, I'll remember it for future documentation updates :)

PR #200, opened 2018-10-07
4-make-products-persistent.rst

Sebra7:
Changed the Position from the Important Box and the XML Language Code


DanielSiepmann:
Thanks @Sebra7

PR #206, opened 2018-11-02
Replace @lazy with doctrine annotation

hliebe:
Because of Feature #83078 the @lazy annotation is deprecated and has to be replaced with @TYPO3\CMS\Extbase\Annotation\ORM\Lazy


mbrodala:
Since we don't really have book versions bound to TYPO3 versions we should add a hint that `@lazy` was used in TYPO3v8 and older.

PR #210, opened 2018-11-14
rewrite LocalizationUtility::translate

franzholz:


mbrodala:
Thanks, that's way better.

PR #211, opened 2018-11-14
escape uses htmlspecialchars

franzholz:


franzholz:
See the file typo3_src-8.7.19/vendor/typo3fluid/fluid/src/Core/Parser/SyntaxTree/EscapingNode.php

    public function evaluate(RenderingContextInterface $renderingContext)
    {
        $evaluated = $this->node->evaluate($renderingContext);
        if (is_string($evaluated) || (is_object($evaluated) && method_exists($evaluated, '__toString'))) {
            return htmlspecialchars((string) $evaluated, ENT_QUOTES);
        }
        return $evaluated;
    }

PR #224, opened 2019-01-14
Corrected example code

MorgusLethe:
=> was throwing a php error. The value must still be inside brackets [], and probably without the comma.


mbrodala:
You are right, the `>` is wrong here. And yes, the trailing comma was intentional and is highly recommended for any array everywhere. This makes it a lot easier to add more items and reduces the diff for Git.

PR #232, opened 2019-03-09
Extending from AbstractValidator

linawolf:
When extending an AbstractValidator isValid($value) has to be implemented as it is abstract. validate($value) automatically calls isValid($value)
See: https://api.typo3.org/typo3cms/current/html/class_t_y_p_o3_1_1_c_m_s_1_1_extbase_1_1_validation_1_1_validator_1_1_abstract_validator.html#a72063d2e922edda5321b1970297eb0c3


alexanderschnitzler:
This PR is right since the code example shows a validator that inherits `AbstractValidator`. While this change is right, I'd like to see a change in the overall docs in the future, showing that the `ValidatorInterface` is the actual important API.

If you merge this, please also merge this into 9 and latest.

PR #254, opened 2019-04-17
include plugin via typoscript

linawolf:
I was feeling like information on how to include the plugin via typoscript was missing here. The code to be used changed from version to version it seems. So there is a lot of information out there which doesnt work anymore in 8LTE and above

### DIFF
--- a/Documentation/4-FirstExtension/7-configuring-the-plugin.rst
+++ b/Documentation/4-FirstExtension/7-configuring-the-plugin.rst
@@ -91,7 +91,18 @@ The next call of the page, with the plugin on it, shows the inventory as a table
    :align: center
 
    Figure 4-5: The output of the inventory in the front end
+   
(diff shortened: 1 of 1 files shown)


Documentation/4-FirstExtension/7-configuring-the-plugin.rst
```diff
@@ -91,7 +91,18 @@ The next call of the page, with the plugin on it, shows the inventory as a table
    :align: center
 
    Figure 4-5: The output of the inventory in the front end
-
+   
+To include the plugin anywhere on your page via typoscript you can use the bootstrap user function:
```
mbrodala: Suggestion:

> To include the plugin globally in your site via TypoScript you can use a `USER` content object and the Extbase bootstrap:

Also I'd move this to the end of the page as a bonus since it's not essential for the basic plugin-as-content case.

PR #258, opened 2019-05-10
Add license information

sypets:


sypets:
Clarification: The issue of license for the book has been raised via email. Sebastian Kurfürst has ok'd this license. I would like to wait for offical confirmation from someone.

sypets:
Talked about topic in Slack Call. License information can be added.

PR #279, opened 2019-07-01
Security mention for identity field

OlivierSC:
Add security mention about manipulation of identity field value to prevent data replacement by an attacker


DanielSiepmann:
Thanks for adding this hint.

Issue #281, opened 2019-07-12
Document Switchable Controller Actions feature


DanielSiepmann:
This knowledge is necessary for becoming a certified developer. Therefore this feature should be documented.

https://my.skilldisplay.eu/skill/289/37

PR #290, opened 2019-07-18
Document switchableControllerActions

sypets:
To give some context, a chapter "Configuration" is added. This gives an
overview over possibilities and links to other manuals where the concepts
are explained in more detail.

Resolves: #281


DanielSiepmann:
We need to backport this to 9.5 and 8.7.

PR #301, opened 2019-08-02
Added some changes

danielsiepmann-testing:


DanielSiepmann:
We won't merge this one. It was for demonstration purposes during dev days. This should be in sync with other manuals and manual title.

Issue #282, opened 2019-07-12
Document annotations


DanielSiepmann:
This is necessary in order to become certified developer. Therefore this should be documented.

Stuff that might be useful while writing:
* https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.3/Feature-83167-ReplaceValidateWithTYPO3CMSExtbaseAnnotationValidate.html
* https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5/Deprecation-85980-InternalAnnotationInExtbaseCommands.html
* https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/7.0/Feature-60822-AddMethodsToGetClassTagValuesViaReflection.html
* https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/7.5/Feature-68746-AddAnnotationForCLIOnlyCommands.html
* https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Feature-83094-ReplaceIgnorevalidationWithTYPO3CMSExtbaseAnnotationIgnoreValidation.html
* https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Feature-83093-ReplaceCascadeWithTYPO3CMSExtbaseAnnotationORMCascade.html
* https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Feature-83092-ReplaceTransientWithTYPO3CMSExtbaseAnnotationORMTransient.html
* https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Feature-83078-ReplaceLazyWithTYPO3CMSExtbaseAnnotationORMLazy.html
* https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Feature-82869-ReplaceInjectWithTYPO3CMSExtbaseAnnotationInject.html

https://my.skilldisplay.eu/skill/550/37

PR #307, opened 2019-08-23
Add annotation reference

DanielSiepmann:
Resolves: #282
Releases: 9.5, master

### DIFF
--- a/Documentation/b-ExtbaseReference/Index.rst
+++ b/Documentation/b-ExtbaseReference/Index.rst
@@ -572,6 +572,157 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
 mappings available. This default `errorAction` refers back to the referrer, if the referrer
 was sent with it.
 
+.. _available-annotations:
+
+Available annotations
+^^^^^^^^^^^^^^^^^^^^^
+
+All available annotations for Extbase are placed within namespace :php:`TYPO3\CMS\Extbase\Annotation`.
+They can be imported into the current namespace, e.g.::
+
+   use TYPO3\CMS\Extbase\Annotation\Inject;
+
+   /**
+    * @Inject
+    * @var Foo
+    */
+   public $property;
+
+Is completely valid and will be parsed. It's considered to be best practice to
+use the following instead, in order to make source of annotation more
+transparent::
+
+   use TYPO3\CMS\Extbase\Annotation as Extbase;
+
+   /**
+    * @Extbase\Inject
+    * @var Foo
+    */
+   public $property;
+
+.. note::
+
+   In TYPO3 CMS 9.5 it's possible to still use deprecated versions of
+   annotations.
+   To guarantee compatibility with future versions,
+   it's recommended to use the non-deprecated annotations,
+   unless it's not possible due to backwards compatibility need.
+
+The following annotations are available out of the box within Extbase:
+
+:php:`@TYPO3\CMS\Extbase\Annotation\Validate`
+   Allows to configure validators for properties and method arguments:
+
+   .. code-block:: php
+
+      /**
+       * Existing TYPO3 validator.
+       *
+       * @Extbase\Validate("EmailAddress")
+       */
+      protected $email = '';
+
+      /**
+       * Existing TYPO3 validator with options.
+       *
+       * @Extbase\Validate("StringLength", options={"minimum": 1, "maximum": 80})
+       */
+      protected $title = '';
+
+      /**
+       * Custom validator identified by FQCN.
+       *
+       * @Extbase\Validate("\Vendor\ExtensionName\Validation\Validator\CustomValidator")
+       */
+      protected $bar;
+
+      /**
+       * Custom Validator identified by dot syntax, with additional parameters.
+       *
+       * @Extbase\Validate("Vendor.ExtensionName:CustomValidator", param="barParam")
+       */
+      public function barAction(string $barParam)
+      {
+          return '';
+      }
+
+   The above list provides all possible references to an validator. Available
+   validators shipped with Extbase can be found within
+   :file:`EXT:extbase/Classes/Validation/Validator/`.
+
+   Legacy version provided :php:`@validate` which is deprecated. :php:`@Extbase\Validate` should be used instead.
+
+
+:php:`@TYPO3\CMS\Extbase\Annotation\IgnoreValidation()`
+   Allows to ignore Extbase default validation for a given argument, in context
+   of an controller.
+
+   .. code-block:: php
+
+      /**
+       * @Extbase\IgnoreValidation("param")
+       */
+      public function method($param)
+      {
+      }
+
+   Legacy version provided :php:`@ignorevalidation $param` which is deprecated.
+   :php:`@Extbase\IgnoreValidation("param")` should be used instead.
+
+:php:`@TYPO3\CMS\Extbase\Annotation\ORM\Cascade("remove")`
+   Allows to remove child entities during deletion of aggregate root.
+
+   .. code-block:: php
+
+      /**
+       * @Extbase\ORM\Cascade("remove")
+       */
+      public $property;
+
+   Legacy version provided :php:`@cascade remove` which is deprecated.
+   :php:`@Extbase\ORM\Cascade("remove")` should be used instead.
+
+:php:`@TYPO3\CMS\Extbase\Annotation\ORM\Transient`
+   Marks property as transient (not persisted).
+
+   .. code-block:: php
+
+      /**
+       * @Extbase\ORM\Transient
+       */
+      public $property;
+
+   Legacy version provided :php:`@transient` which is deprecated.
+   :php:`@Extbase\ORM\Transient` should be used instead.
+
+:php:`@TYPO3\CMS\Extbase\Annotation\ORM\Lazy`
+   Marks property to be lazy loaded, on first access.
+
+   .. code-block:: php
+
+      /**
+       * @Extbase\ORM\Lazy
+       */
+      public $property;
+
+   Legacy version provided :php:`@lazy` which is deprecated.
+   :php:`@Extbase\ORM\Lazy` should be used instead.
+
+:php:`@TYPO3\CMS\Extbase\Annotation\Inject`
+   Configures dependency injection (DI) to inject given property.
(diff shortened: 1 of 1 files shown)


Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+
+   /**
+    * @Extbase\Inject
+    * @var Foo
+    */
+   public $property;
+
+.. note::
+
+   During TYPO3 CMS 9.5 it's possible to still use deprecated versions of
```
susannemoog: "In" or "Using" instead of "during" as TYPO3 CMS 9.5 is no time frame.

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+
+   use TYPO3\CMS\Extbase\Annotation\Inject;
+
+   /**
+    * @Inject
+    * @var Foo
+    */
+   public $property;
+
+Is completely valid and will be parsed. It's considered to be best practice to
```
susannemoog: Maybe add a half-sentence as to the reasoning (more transparent to distinguish extbase from other annotations)

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+   /**
+    * @Extbase\Inject
+    * @var Foo
+    */
+   public $property;
+
+.. note::
+
+   During TYPO3 CMS 9.5 it's possible to still use deprecated versions of
+   annotations. Due to future compatibility it's recommended to use up to date
```
susannemoog: This sentence sounds weird. Maybe change to: `To guarantee compatibility with future versions, it's recommended to use the non-deprecated annotations.` (if we need that sentence at all, because it's kind of what a deprecation is)

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+   annotation, until it's not possible due to backwards compatibility need.
+
+The following annotations are available out of the box within Extbase:
+
+:php:`@TYPO3\CMS\Extbase\Annotation\Validate`
+   Allows to configure validators for properties and method arguments:
```
susannemoog: do we have a list of available validators we could link?

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+   .. code-block:: php
+
+      /**
+       * @Extbase\IgnoreValidation("param")
+       */
+      public function method($param)
+      {
+      }
+
+   Legacy version used :php:`@ignorevalidation $param` instead of :php:`@Extbase\IgnoreValidation("param")`.
```
susannemoog: Maybe: "Legacy version provided ... which is deprecated. ... should be used instead.

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+
+      /**
+       * @Extbase\ORM\Cascade("remove")
+       */
+      public $property;
+
+   Legacy version used :php:`@cascade remove` instead of :php:`@Extbase\ORM\Cascade("remove")`.
+
+:php:`@TYPO3\CMS\Extbase\Annotation\ORM\Transient`
+   Marks property to be transient (not persisted).
```
susannemoog: "Marks property as transient"

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+
+      /**
+       * @Extbase\ORM\Transient
+       */
+      public $property;
+
+   Legacy version used :php:`@transient` instead of :php:`@Extbase\ORM\Transient`.
+
+:php:`@TYPO3\CMS\Extbase\Annotation\ORM\Lazy`
+   Marks property to be lazy loaded, on first access.
```
susannemoog: Do we need to add more explanation to transient or lazy or is that written down somewhere else?

Documentation/b-ExtbaseReference/Index.rst
alexanderschnitzler: The sentence makes much more sense if you replace `until` with `unless`. Still, I wouldn't really state, that there are legacy versions of those annotations available. People who come from previous versions know this already and those who start with 9.5+ don't need to learn about the past.

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+       *
+       * @Extbase\Validate("\TYPO3\CMS\Extbase\Tests\Functional\Validation\Fixture\Validation\Validator\CustomValidator")
+       */
+      protected $bar;
+
+      /**
+       * Custom Validator identified by FQCN using dot syntax, with additional
+         parameters.
+       *
+       * @Extbase\Validate("TYPO3.CMS.Extbase.Tests.Functional.Mvc.Controller.Fixture:CustomValidator", param="barParam")
```
alexanderschnitzler: This syntax is really just relevant in it's short form IMHO. Using a FQCN with dots instead of backslashes raises the question why this syntax option exists in the first place.

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+      public function barAction(string $barParam)
+      {
+          return '';
+      }
+
+   Legacy version used :php:`@validate` instead of :php:`@Extbase\Validate`.
+
+
+:php:`@TYPO3\CMS\Extbase\Annotation\IgnoreValidation()`
+   Allows to ignore Extbase default validation for a given argument.
```
alexanderschnitzler: We should state that the context is a controller action here. The `IgnoreValidation` annotation is evaluated no matter the context, but the `Validate` annotations are bound to controller actions, so it doesn't make sense using `IgnoreValidation` anywhere else and thus we can state that those are intended to be used with controller actions.

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+
+      /**
+       * @Extbase\ORM\Lazy
+       */
+      public $property;
+
+   Legacy version used :php:`@lazy` instead of :php:`@Extbase\ORM\Lazy`.
+
+:php:`@TYPO3\CMS\Extbase\Annotation\Inject`
+   Configured dependency injection (DI) to inject given property.
```
alexanderschnitzler: A general question about tenses. Do we write present or past tense? Because currently it's mixed.
`Configured` vs. `Configures`

Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -572,6 +572,144 @@ in `$this->argumentsMappingResults` you have a list of occurred warnings and err
...
+
+   .. code-block:: php
+
+      /**
+       * @Extbase\Inject
+       * @var TypeAnnotation
```
alexanderschnitzler: Maybe replace `TypeAnnotation` with `ServiceClassName` to clarify things. And if we do this, we should distinct between version 9.5 and 10.0 as one has to use FQCN's until 9.5.

Documentation/b-ExtbaseReference/Index.rst
DanielSiepmann: Thanks. The reason I've added the legacy stuff to 9.5 only (not part of master) is that this might be asked during certification of 9 LTS:

>>> 
    I can use the correct Extbase annotations in my models and controllers
    I can decide which annotation is the correct to use when there are several possible solutions

Documentation/b-ExtbaseReference/Index.rst
DanielSiepmann: I'll add the folder where existing Validators are placed.

Documentation/b-ExtbaseReference/Index.rst
DanielSiepmann: That is explained at https://docs.typo3.org/m/typo3/book-extbasefluid/master/en-us/5-Domain/2-implementing-the-domain-model.html#implementing-relationships-between-domain-objects

Documentation/b-ExtbaseReference/Index.rst
DanielSiepmann: Do we have an example for short version? I've copied this from tests at https://github.com/TYPO3/TYPO3.CMS/blob/9.5/typo3/sysext/extbase/Tests/Functional/Mvc/Controller/Fixture/Controller/TestController.php#L53

Documentation/b-ExtbaseReference/Index.rst
DanielSiepmann: I would like to use present, sorry for mixing things up.

But as this is current state, present would make sense.

Documentation/b-ExtbaseReference/Index.rst
DanielSiepmann: These hints are also not part of PR into `master`. Therefore they only exist in 9.5 context, where they still exist as deprecation.

Documentation/b-ExtbaseReference/Index.rst
alexanderschnitzler: You can find a couple of different versions here:
https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/extbase/Tests/Unit/Reflection/Fixture/DummyModel.php#L31

I'd say, either use a real FQCN for custom validators and the ultra short form `NotEmpty` e.g. for all Extbase validators.

PR #309, opened 2019-09-21
Add docker compose file

janmaennig:
That should be made renderable this part of documentation out of the box. I think, for a contribution newbee like ne, it is helpfull that every part is rendering out of the box. If that is desired, I can do the same for the other repos.


janmaennig:
Hello marble! I will pack the file into a PR in the next few days. Ansich I had already taken her from one of the Docs Repos anyway.

PR #310, opened 2019-09-23
Feature/#114 localization arguments

janmaennig:
Add an example to use placeholder respectively wildcards in translation.

### DIFF
--- a/Documentation/b-ExtbaseReference/Index.rst
+++ b/Documentation/b-ExtbaseReference/Index.rst
@@ -953,7 +953,7 @@ does all the translation. The method can be called like this:
     The extension name. It can be fetched from the request.
 
 `$arguments`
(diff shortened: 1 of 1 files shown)


Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -953,5 +953,31 @@ does all the translation. The method can be called like this:
     The extension name. It can be fetched from the request.
 
 `$arguments`
-    Allows you to specify an array of arguments passed to the function `vsprintf`. Allows you to
-    fill wildcards in localized strings with values.
+    Allows you to specify an array of arguments. In the `LocalizationUtility` this arguments will be passed to the function `vsprintf`. So you can insert dynamic values in every translation. You can find the possible wildcard specifiers under `https://www.php.net/manual/en/function.sprintf.php#parameter <https://www.php.net/manual/en/function.sprintf.php#parameter>`__.
```
DanielSiepmann: Pro tip: Do not include `/en` within php.net urls. php.net will then detect your current language and redirect. This allows people to use the documentation in their preferred language.

DanielSiepmann: If URL and text for link are the same, there is no need to duplicate them, just remove ` <url>` part.

janmaennig: Thanks ... I have updated the link.

PR #312, opened 2019-09-27
Fix typo in 5-controlling-the-flow.rst

kenchoihm:
I believe it is a typo in the documentation ([https://docs.typo3.org/m/typo3/book-extbasefluid/master/en-us/4-FirstExtension/5-controlling-the-flow.html]), and I think the author meant "defined" (instead of "declined").


DanielSiepmann:
Thanks, good catch :+1:

PR #333, opened 2020-01-22
Improve readability of Validating Domain Objects page

tomwarwick:
This commit intends to improve the overall readability of the Validating Domain Objects page.

- Improve overall readability making it easier to read and hopefully understand.
- Remove any uses of the phrase "so called" when introducing the reader to new items
- Capitalise the word Extbase

### DIFF
--- a/Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
+++ b/Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
@@ -279,7 +279,7 @@ objects are validated.
 
 
 Validating in the domain model with your own validator class
-----------------------------------------------------------
+------------------------------------------------------------
 
 The ability to register validators in the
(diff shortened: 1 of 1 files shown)


Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
```diff
@@ -278,12 +278,12 @@ the domain model. The next section shows how complex domain
 objects are validated.
 
 
-Validating in the domain model with an own validator class
+Validating in the domain model with your own validator class
```
sypets: Remember that the underline must always be at least as long as the title / section header. If not, this may not get rendered correctly or produce warnings. 

I can push another commit to fix this.

sypets:
Apart from that, these look like great changes and I will try to get this merged as quickly as possible.

Thanks for taking the time! :thumbsup:

PR #337, opened 2020-03-15
Changes in wording

sypets:
Commits from @ thombansen

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "restructuredtext.confPath": ""
+}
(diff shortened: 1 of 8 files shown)


thombansen:
Thanks for troubleshooting this. True - I edited the 9,5 version as I thought it would relate to the version that i was looking for. Apologies.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
sypets, line 1: .vscode/settings.json is an IDE specific configuration. I am not sure if we should include those in the documentation repositories in general. If we do though, we should add these to all repositories. Please make a suggestion as issue in "Writing Documentation" (were we documentation the general conventions): https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/issues/new

I assume you could also add your settings in your home directory instead.

I will remove this file from the PR.

PR #347, opened 2020-06-19
Update Settings.cfg

126016:
Add chapter4 link in Settings.cfg

PR #348, opened 2020-06-19
Add a link target for "chapter 4"

126016:
Add a link target for "chapter 4"

### DIFF
--- a/Documentation/3-BlogExample/Index.rst
+++ b/Documentation/3-BlogExample/Index.rst
@@ -23,7 +23,7 @@ background. This knowledge will then help you in the creation of your own
 extension.
 
 If you are looking for a manual specifically on creating an extension,
-:ref:`chapter 4 <creating-a-first-extension>` will show you the right set of tools. However, we recommend to
+:ref:`creating-a-first-extension` will show you the right set of tools. However, we recommend to
 work on the fundamentals in this chapter. The journey that lies ahead
 of us could also have the title "Europe in five days." If you discover nice
 places, you should visit them later without the travel group.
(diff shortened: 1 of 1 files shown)


Documentation/3-BlogExample/Index.rst
```diff
@@ -23,7 +23,7 @@ background. This knowledge will then help you in the creation of your own
 extension.
 
 If you are looking for a manual specifically on creating an extension,
-chapter 4 will show you the right set of tools. However, we recommend to
+:ref:`chapter 4 <creating-a-first-extension>` will show you the right set of tools. However, we recommend to
```
mbrodala, line 26: Since you mentioned this in #347: please use this instead: 

```
:ref:`creating-a-first-extension`
```

126016: Okay 
:ref:`creating-a-first-extension` is the syntax for the link, but where come the text for the link "chapter 4" ?

126016: I found this documentation:
https://www.sphinx-doc.org/en/1.7/markup/inline.html#cross-referencing-arbitrary-locations

And it says that "Labels that aren’t placed before a section title can still be referenced, but you must give the link an explicit title, using this syntax: :ref:`Link title <label-name>`." 

So I used that. Not quite sure I am doing this for the first time.

mbrodala: It is inserted automatically, see https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingReST/CheatSheet.html#cheat-sheet-intersphinx

126016: I change it in :ref:`creating-a-first-extension`

mbrodala:
@126016 the change has been [deployed now](https://docs.typo3.org/m/typo3/book-extbasefluid/master/en-us/3-BlogExample/Index.html):

image

PR #352, opened 2020-06-27
Fix broken link to cheat seets

fawmi:


Documentation/b-ExtbaseReference/Index.rst
```diff
@@ -11,7 +11,7 @@ Extbase extensions.
 
 .. note::
 
-    Under https://docs.typo3.org/typo3cms/CheatSheets.html you will find
+    Under https://docs.typo3.org/m/typo3/docs-cheatsheets/master/en-us/CheatSheets.html#cheat-sheets you will find
```
sypets, line 14: The cheatsheets do not contain an Extbase / Fluid cheatsheet anymore. I think this is only available for older versions (e.g. v 7.6). So I would remove this link entirely.

PR #353, opened 2020-07-30
Use correct ViewHelpers

dakoether:
If I am right there is no f.form.textbox Viewhelper. It should be f.form.textfield for "title" and f.form.textarea for "description" instead.


sypets:
Thank you for your contribution! :thumbsup:

PR #366, opened 2020-10-17
[TASK] Remove switchableControllerActions

susannemoog:
See Core Deprecation 89463
See https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/845


derhansen (approved):

mbrodala (approved):

mbrodala:
Normally I'd expect a note here but in this case it's probably fine to drop without replacement since adding plugins itself is already covered.

PR #374, opened 2020-11-30
[TASK] Extbase controller actions must return ResponseInterface

brotkrueml:
See: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Deprecation-92784-ExtbaseControllerActionsMustReturnResponseInterface.html

Missing: How to handle redirect() and forward() calls inside actions?


susannemoog:
I did some changes for this change already: be careful with the examples code as that is taken directly Form the blog exanple extension and verified to be working. Especially: the actions of a controller extending the action controller may still return a string as that is converted to a proper response object by the parent as far as I understood.

susannemoog:
> I did some changes for this change already: be careful with the examples code as that is taken directly Form the blog exanple extension and verified to be working. Especially: the actions of a controller extending the action controller may still return a string as that is converted to a proper response object by the parent as far as I understood.

To clarify: that is deprecated, but we should adjust the blog example extension first and then update the docs here, otherwise we get a mismatch between the two.

brotkrueml:
> To clarify: that is deprecated, but we should adjust the blog example extension first and then update the docs here, otherwise we get a mismatch between the two.

I agree, will have a look at the extension.

Documentation/3-BlogExample/4-and-action.rst
```diff
@@ -14,64 +14,83 @@ In Extbase the controllers mostly exist as
...
+      }
+
+      public function newAction(Blog $newBlog = null): ResponseInterface
+      {
+         $this->view->assignMultiple([
+            'newBlog' => $newBlog,
+            'administrators' => $this->administratorRepository->findAll(),
+         });
+
+         return $this->responseFactory->createHtmlResponse($this->view->render());
```
alexanderschnitzler, line 55: There is helper method `htmlResponse()` to ease the migration. We should use that in the blog example but explain this method and using the responseFactory while mentioning PSR-17.

Documentation/3-BlogExample/4-and-action.rst
```diff
@@ -14,64 +14,83 @@ In Extbase the controllers mostly exist as
...
+      }
+   }
 
 The method `indexAction()` within the
 :php:`BlogController` is responsible for showing a list of
 blogs. We also could have called it
 `showMeTheListAction()`. The only important point is,
 that it ends with `Action` in order to help Extbase
-to recognize it as an action. `newAction()` shows a
+to recognize it as an action and returns an implementation of the
```
alexanderschnitzler, line 92: With this change the sentence is grammatically incorrect.

Documentation/3-BlogExample/8-Back-in-the-controller.rst
```diff
@@ -32,41 +32,39 @@ The following example explains this mechanism:
...
-       public function fooAction()
-       {
-           // foo
-       }
-
-       public function barAction()
-       {
-           // bar
-       }
+      public function initializeAction()
```
alexanderschnitzler, line 44: Can we please just change one thing at a time, whitespaces or logic stuff? It makes the diff much easier.

Documentation/3-BlogExample/8-Back-in-the-controller.rst
```diff
@@ -32,41 +32,39 @@ The following example explains this mechanism:
...
+      {
+         // bar
+      }
    }
 
-Action methods can either return a response (`string` or object of type `\Psr\Http\Message\ResponseInterface`)
-themselves or not return at all.
-If an action method does not return anything, Extbase calls `$this->view->render()` and
-returns the rendered view automatically.
+Action methods return a response object of type `\Psr\Http\Message\ResponseInterface`.
```
alexanderschnitzler, line 66: Please a the word `must` here.

alexanderschnitzler (changes requested):
Most importantly, please revert the whitespace changes and all other changes that do not deal with this explicit change in Extbase. Think of it this way: You encounter a change in Extbase, read the docs, lookup the commits and then see a diff with dozens of changes instead of just the necessary change. That's quite distracting.

brotkrueml:
@alexanderschnitzler Will take care of your suggestions. Before further processing of this issue I started to cleanup the blog_example code and then want to begin with adapting for v11. But the first PRs are not merged yet.

linawolf:
What is the status of this PR, @brotkrueml ?

linawolf:
sorry, accidently hit the wrong button...

brotkrueml:
@linawolf I started pull requests on the blog_example extension, but there was no feedback since. The goal is to make the blog_example extension clean, then make it v11 ready. After that adopt the code examples to the documentation.

linawolf:
So the question would be how to succeed.. Is there anyone who we could ping?

susannemoog:
@brotkrueml / @linawolf I checked the open PRs on the blog_example and merged them, so work can continue. Additionally, I gave you - Lina - write access to the blog example repository.

linawolf:
So @brotkrueml is it possible to merge this now or do you need further changes?

brotkrueml:
Hmm, don't know. The blog_example extension should be in line with the examples in this docs, as mentioned in a comment above. As this should be no problem, the tags doesn't work for me (in 10.4). Wanted to investigate that, but had other priorities. 

I was enthusiastic when the change was merged into the v11 core. Then started to dig into the blog example extension, adjusted some minor issues also for 10.4 before making the adjustments for v11. But the merges lasts some weeks and so my priorities shifted.

From my side, it can be merged and later make more adjustments. Someone can also take over the PR or it can be closed and make a fresh start.

linawolf (approved):

linawolf (approved):

PR #386, opened 2020-12-30
Set recommended line length to 80

linawolf:
https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/CodingGuidelines.html?highlight=editorconfig


linawolf:
It was decided not to do this

PR #387, opened 2021-01-19
change $_EXTKEY to extension string

Teisi:
cause $_EXTKEY is obsolete: https://wiki.typo3.org/Exception/CMS/1365429656


mbrodala:
Yes, it is obsolete everywhere *except* in `ext_emconf.php`. Here you *must* continue to use it, otherwise package installation can break: https://github.com/TYPO3/CmsComposerInstallers/pull/59

mbrodala:
I've just added this to the wiki page.

PR #389, opened 2021-01-28
add renderType definition to TCA language fields

mcmietz:
It seems, that TCA type select need a renderType definition. 
Otherwise you got "Unknown type: select, render type: select" in Backend View in TYPO3 10.4.12


mbrodala (approved):

mbrodala:
Yes, absolutely correct, thanks for the fix!

PR #390, opened 2021-01-28
change naming off TCA field for language parameter

mcmietz:
In the paragraph before code example fields "l10n_parent" "l10n_diffsource" are dicribed.
But in code example "l18n_parent" and "l18n_diffsource" is used. But this fields only used for tt_content table. 
Changed naming of TCA fields to "l10n_parent" and "l10n_diffsource".


mcmietz:
Theres something missing, Create another one

PR #391, opened 2021-01-28
Delete spaces in TCA example

mcmietz:
If you copy the code, the code is added in more than one line. 
This could lead to errors and confusion if there is no error output. Especially for beginners


susannemoog:
will do backports and formatting fixes in whole file

PR #392, opened 2021-01-28
change naming off TCA field for language parameter

mcmietz:
In paragraph before TCA code example fields "l10n_parent" "l10n_diffsource" are dicribed.
But in code example "l18n_parent" and "l18n_diffsource" is used. But this fields only used for tt_content table.
Changed naming of TCA fields to "l10n_parent" and "l10n_diffsource".

### DIFF
--- a/Documentation/9-CrosscuttingConcerns/1-localizing-and-internationalizing-an-extension.rst
+++ b/Documentation/9-CrosscuttingConcerns/1-localizing-and-internationalizing-an-extension.rst
@@ -341,7 +341,7 @@ translation relates to.
    return [
        // ...
        'types' => [
-           '1' => ['showitem' => 'l18n_parent , sys_language_uid, hidden, title,
+           '1' => ['showitem' => 'l10n_parent , sys_language_uid, hidden, title,
                        description, logo, posts, administrator'],
        ],
        'columns' => [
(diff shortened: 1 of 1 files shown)


susannemoog:
I'm never sure which one is technically more correct - can someone check with the original code in blog example? (this code should match the code there) From the word itself (localization) - l10n would be correct.

mbrodala:
@susannemoog Exactly. That's why `l18n` does not make sense. This is a mix of `l10n` (`localization`) and `i18n` (`internationalization`).

So again: yes, `l10n_` must be used.

mcmietz:
Ok, so the blog example really creates fields l18n_parent and l18n_diffsource in the database for table tx_blogexample_domain_model_blog. 
For person, again, l10n_ .

Should we rather smooth this out in the blog example and then adjust the doc accordingly?

mbrodala:
Yes, this would make sense IMO. We should not keep broken examples around for historic reasons. It's bad enough that we have a popular example with `tt_content` which still does it wrong.

linawolf:
@mbrodala @mcmietz What is the status here? Can it be merged?

mcmietz:
@linawolf I think it's waiting for the pull request in the blog example, I wouldn't merge the change until it's changed there, otherwise the explanation differs from the example code.

susannemoog:
Pull request in blog_example is merged.

PR #403, opened 2021-02-23
Added link to fluidconverter

### LINKED ISSUE #219: Clarify Fluid Inline Notation
see https://docs.typo3.org/typo3cms/ExtbaseFluidBook/8-Fluid/1-basic-concepts.html#inline-notification-for-view-helpers

* should be "inline notation", not "Inline Notification"?
* this chapter is hidden, perhaps move it up in toc hierarchy
* possibly, add reference to http://www.fluid-converter.com/

linawolf:
Releases master, 10.4
@resolves #219

### DIFF
--- a/Documentation/8-Fluid/1a-inline-syntax.rst
+++ b/Documentation/8-Fluid/1a-inline-syntax.rst
@@ -6,6 +6,11 @@
 Inline notation for ViewHelpers
 ===============================
 
+.. tip::
+
+   There is an online converter from tag-based syntax to inline syntax:
+   `Fluid Converter <http://www.fluid-converter.com/>`__
+
 It is intuitive and natural for most of the ViewHelpers to be called
 with the tag-based syntax. Especially with control structures or form
 elements, this syntax is easily understood. But there are also ViewHelpers,


Documentation/8-Fluid/1a-inline-syntax.rst
```diff
@@ -6,6 +6,11 @@
 Inline notation for ViewHelpers
 ===============================
 
+.. tip::
+
+   There is an online converter from tag-based syntax to inline syntax:
+   `Fluid Converter <http://www.fluid-converter.com/>`__
```
mbrodala, line 12: Before we add a link to this, we should try to have it use HTTPS instead. I've created a ticket: https://github.com/stat1x/fluid-converter/issues/11

linawolf: Why should this keep us from merging? The link can always be changed in the manual later on.

mbrodala: You are right.

mbrodala (approved):

PR #415, opened 2021-03-12
no "we"

franzholz:


Documentation/4-FirstExtension/1-the-example-extension.rst
```diff
@@ -7,8 +7,8 @@
 The example extension
 =====================
 
-Our first extension will show an inventory list of products,
-which we created in a backend list module.
+The first extension will show an inventory list of products,
+which can be created in the backend list module.
```
mbrodala, line 11:
```suggestion
which is maintained in the backend *List* module.
```

This hopefully makes it clear that not the extension is created in the backend List module.

Documentation/4-FirstExtension/1-the-example-extension.rst
```diff
@@ -19,10 +19,10 @@ The following steps are necessary for implementation:
 
 
    * Define the database tables
-   * Configure the display of backend forms
-   * Create repositories for product objects
+   * Configure the display of the backend forms
+   * Create the repository for the product object
```
mbrodala, line 23: Actually `objects` is correct here since this is barely about a single instance.

Alternatively use something like _`Product` type_.

mbrodala (changes requested):

linawolf (approved):

PR #444, opened 2021-05-12
"first extension" is a bad title

franzholz:
There has already been the blog example as the first extension.

### DIFF
--- a/Documentation/4-FirstExtension/1-the-example-extension.rst
+++ b/Documentation/4-FirstExtension/1-the-example-extension.rst
@@ -3,11 +3,11 @@
 
 .. include:: /Includes.rst.txt
 
-=====================
-The example extension
-=====================
+=============================
+The store inventory extension
+=============================
 
-The first extension will show an inventory list of products,
+An inventory list of products is shown as an example extension,
 which is maintained in the backend *List* module.
 Each product consists of a title,
 a short description, and the number of pieces in stock.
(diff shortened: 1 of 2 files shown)


Documentation/4-FirstExtension/1-the-example-extension.rst
```diff
@@ -3,11 +3,11 @@
 
 .. include:: /Includes.rst.txt
 
-=====================
-The example extension
-=====================
+=============================
+The store inventory extension
```
DavidBruchmann, line 7: In general I think the headline can reflect somehow the difficulty, this is lost when it's getting just the extension name which doesn't reflect this.

franzholz: There is more than one example extension. I always think about the blog example as the first extension. So this title is misleading and not helpful.

DavidBruchmann: sure, the blog is historically well-known, but to newbies perhaps not.  
Anyway, if you think my comment is not useful, just feel free to ignore it.

linawolf (approved):

Documentation/4-FirstExtension/1-the-example-extension.rst
linawolf: I would agree with @franzholz as there is more then one example extension naming the concrete extension makes navigating easier.

PR #448, opened 2021-05-15
missing information for redirect

franzholz:


Documentation/10-Outlook/6-dispatching.rst
```diff
@@ -283,16 +286,30 @@ returning a `ForwardResponse`::
...
+   
    $this->redirectToUri('https://example.com');
 
+.. note::
+
+   A redirection leads to a reload of the page. All the $_REQUEST variable is lost. Therefore all data needed on the 
+   destination must be passed as an array in parameter 4. 
+   If the redirection happens after a new / create form, then it mudst be taken care that the database record is stored
+   into the database before the redirection:
+   `$this->objectManager->get(\TYPO3\CMS\Extbase\Persistence\Generic\PersistenceManager::class)->persistAll();`
```
susannemoog, line 303: the object manager is deprecated and should not be used anymore - we should not add it to new documentation

franzholz: The ObjectManager has been deprecated in TYPO3 11.2. But this is no published TYPO3 yet. And it still works in TYPO3 11.2.
I have added a TODO.

susannemoog: Your PR targets "master" which is TYPO3 v11-dev and 11.2 has actually been released and is a published version, that's why we should not use the object manager here.

franzholz: But this Object Manager part is needed for the document versions for TYPO3 9 and 10.

linawolf: @franzholz please rebase your PR to LTS 10 if it is needed for that version and below only

franzholz: Most changes are needed also in TYPO3 11. 
Only a few lines are still working, but already outdated in TYPO3 11.

linawolf (approved):

PR #452, opened 2021-05-26
There is no document "validation API".

franzholz:
See Slack Documentation from yesterday. Simon Gilli

### DIFF
--- a/Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
+++ b/Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
@@ -46,6 +46,33 @@ The interface requires validators to implement two methods:
 - :php:`validate($value)`
 - :php:`getOptions()`
 
+::
+
+    namespace TYPO3\CMS\Extbase\Validation\Validator;
+
+    /**
+    * Contract for a validator
+    */
+    interface ValidatorInterface
+    {
+        /**
+        * Checks if the given value is valid according to the validator, and returns
+        * an Error Result object.
+        *
+        * @param mixed $value The value that should be validated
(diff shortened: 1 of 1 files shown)


Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
```diff
@@ -69,8 +69,7 @@ value against.
 
 .. tip::
 
-    You will find the complete reference of the
-    :php:`ValidatorInterface` in the :ref:`validation-api`.
+    The complete API of the :php:`ValidatorInterface` is `here https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/extbase/Classes/Validation/Validator/ValidatorInterface.php`__.
```
mbrodala, line 72: We should not link to files in the source since these are bound to be removed, renamed, moved, etc. at any point.

If at all we should link to a specific revision since `master` is symbolic (and can thus change) whereas a commit is fixed.

mbrodala (changes requested):

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
franzholz: TYPO3 is unlikely to be removed from Github.
The link can be adapted again if necessary.

mbrodala: I didn't mean TYPO3 but this file (or any file in the TYPO3 source code).

mbrodala:
TBH we should just drop the whole tip since it doesn't add any value. The full API of `ValidatorInterface` (two methods) is already mentioned at the beginning of this section.

PR #459, opened 2021-06-18
validator modifications

franzholz:


Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
```diff
@@ -72,28 +72,28 @@ This interface requires validators to implement two methods:
 - :php:`getOptions()`
 
 The main method is :php:`validate()`, which is called by the framework.
-The value which is to be validated is passed along to the said method, and it is the
-validator's job to check if that value is valid.
+The value, which is to be validated, is passed along to the said method. It is the
+validator's job to check, if that value is valid.
```
flicstar, line 76:
```suggestion
The value that is to be validated is passed along to the said method. It is the
validator's job to check if that value is valid.
```
If we change "which" to "that" we don't need some of the commas.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
```diff
@@ -72,28 +72,28 @@ This interface requires validators to implement two methods:
...
 
 .. note::
 
     Although the interface states, that the method `validate` should return
-    a :php:`\TYPO3\CMS\Extbase\Error\Result` object, it is not a common practice to do
+    a :php:`\TYPO3\CMS\Extbase\Error\Result` object, it is a common practice not to do
```
flicstar, line 81:
```suggestion
    a :php:`\TYPO3\CMS\Extbase\Error\Result` object, it is not a common practice to do
```
I prefer the original.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
```diff
@@ -72,28 +72,28 @@ This interface requires validators to implement two methods:
...
+    framework. 
 
 If the logic of your validator allows for loose/variable validation checks,
 validator options might come in handy. Extbase ships with a :php:`StringLength`
-validator which offers the options `minimum` and `maximum` that
-let you define the string length the validator should use to check the incoming
-value against.
+validator, which offers the options `minimum` and `maximum`, that
+let you define the string length the validator should use to check against the incoming
+value.
```
flicstar, line 93:
```suggestion
let you define the string length the validator should use to check the incoming
value against.
```
I prefer the original.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
```diff
@@ -72,28 +72,28 @@ This interface requires validators to implement two methods:
...
 
-For example, a validator that checks whether the passed string is
-a valid email address looks like this:
+For example, a validator that checks, whether the passed string is
```
flicstar, line 95:
```suggestion
For example, a validator that checks whether the passed string is
```
not needed

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
```diff
@@ -148,7 +148,7 @@ When they get inserted into a controller action. Figure
 
     Figure 9-1: Data flow of a request before the action is called
 
-When a user sends a request, Extbase first determines which action
+When a user sends a request, Extbase first determines, which action
```
flicstar, line 151:
```suggestion
When a user sends a request, Extbase first determines which action
```
Not needed

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
```diff
@@ -411,8 +426,8 @@ for example:
       }
    }
 
-Now there are two possibilities how validators can be
-registered for domain objects:
+Now there are two possibilities, how validators can be
```
flicstar, line 429:
```suggestion
Now there are two possibilities how validators can be
```
Not needed

flicstar (approved):
Looks good 👍

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
franzholz: Is this a common practice or not?

When something is not  a common practice, why do you mention it at all?

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
franzholz: I prefer to stick to a valid dictionary.

https://www.dict.cc/?s=to+check+against

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
DavidBruchmann: It's a contrast of a solution recommended by the interface versus an alternative with further options.
I think it's quite useful to mention about it because without that explanation they'd be confused about the difference between theory and praxis.

franzholz: So can you answer my question?

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
DavidBruchmann: The value to be validated is passed to the :php:`validate()` method where it's validity is checked.

DavidBruchmann: Keeps perhaps to mention if the return-values of `validate()` are always bool only.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
franzholz: Then my correction of the former text must be correct, or can you explain it?

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
franzholz: It is not of type bool.
See my new propopsal.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
DavidBruchmann: Actually the location where you place the `not` is expressing it's strength.
So if you say `it is a common practice not to do ...`  that means that it's rarely done and most people avoid that solution.
If you say `it is not a common practice to do` leaves it  more open, some use the solution perhaps, some not.
I'd stick to the original version like @flicstar proposed, then you never have to prove anything and the solution is not completely dis-recommended.

franzholz: I do not understand what you mean with "rarely using this solution". Which solution do you mean? 

1. Solution to return a Response object
2. Solution to return a void.

Well, but did you have a look at many TYPO3 extensions and check, what the programmers are using?

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
```diff
@@ -72,28 +72,28 @@ This interface requires validators to implement two methods:
...
+    a :php:`\TYPO3\CMS\Extbase\Error\Result` object, it is a common practice not to do
     so, because most people who create custom validators extend the class
     :php:`\TYPO3\CMS\Extbase\Validation\Validator\AbstractValidator`.
 
     This enables you to call the addError()` method and let the abstract
     validator take care of returning a proper result object to the validation
-    framework.
+    framework. 
 
 If the logic of your validator allows for loose/variable validation checks,
```
DavidBruchmann, line 92: What does this sentence mean exactly?
I think it's a terrible sentence.

franzholz: This is an outdated version of the text.
Read my latest test.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
DavidBruchmann: Yes I remarked, that it's more complicated and seemingly not predefined by an interface.

franzholz: Because it is complicated and hard to understand, it should be well documented.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
DavidBruchmann: I refer to the sentence itself, no matter in which version.
So the solution to use a Response object is not very common.  

I never searched in extension for the common kind of usage, so I don't know which solution is used more often or if the void-solution is used at all.

franzholz: I oppose to your explanation.
You wrote:

>>    So if you say it is a common practice not to do ... that means that it's rarely done and most people avoid that solution.

This is wrong.
If I say that something is a common practice, then it means, that it is frequently done. 

See the English dictionary:

https://www.dict.cc/?s=common+practice

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
DavidBruchmann: for non-native English readers Franz' sentence is probably easier readable.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
DavidBruchmann: Ok, it's perhaps useful to add a sentence to explain the options concerning return-values explicitly?

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
franzholz: I do not know either what a loose validation check is. 
And why is it an alternative to a variable check?

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
DavidBruchmann: `frequently done` can mean that 20% of programmers do it always or 80%, that is completely open.  
Without knowing more detailed which solution is used how often I think the weight is like I explained, just by the location of the `not`. Im speaking mostly English since 8 years, and it's just my feeling concerning both variations.

franzholz: You speak English, but not in England or in the USA. So you still could speak wrong English.

Just show me a reference to a dictionary where this term is explained.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
DavidBruchmann: A loose check is making only a relaxed check, in contrast a tight check requires exact fitting of the value.
I just googled it, so it wasn't clear for me before.
Anyway, the sentence should be changed I think.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
DavidBruchmann: sorry, I won't search for it

DavidBruchmann: beside that placing the not here or there doesn't change any meaning of the term itself.

franzholz: I am sure that  2 different meanings exist.

1. It is a common sense not to do something.
2. It is not a common sense to do something.

The sentences are not the same, but the opposite of each othere.

franzholz: Example:

1. It is a common sense not to walk in the center of a road.
==> Nobody will do it.
2. It is not a common sense to never walk on the side of a road. 
==> Everybody can do it.

Documentation/9-CrosscuttingConcerns/2-validating-domain-objects.rst
```diff
@@ -89,8 +89,12 @@ The value to be validated is passed to the :php:validate() method where it's val
     of type `\TYPO3\CMS\Extbase\Error\Result` to the validation
     framework. 
 
-If the logic of your validator allows for loose/variable validation checks,
-validator options might come in handy. Extbase ships with a :php:`StringLength`
+There are 2 types of validation checks:
+#.  A loose check is making only a relaxed check over a very large range of possible values.
+#.  A tight check requires the exact fitting of the value.
+
+If the logic of your validator allows for loose variable validation checks,
```
DavidBruchmann, line 96: If your validator shall allow a loose variable validation, ...

DavidBruchmann: if you want to implement a loose variable validation, ...

DavidBruchmann: in fact not the variable is checked, but the value.
So the word variable can be removed from my propositions.

PR #468, opened 2021-07-30
[BUGFIX] Fix broken rendering because of inline TODO marker

NeoBlack:


susannemoog (approved):

susannemoog:
todo marker only existed in master, no backport required.

PR #469, opened 2021-07-30
[DEPRECATION]  #94619 - Extbase ObjectManager

linawolf:
refs https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/1449


susannemoog:
We should remove all other usages and references to the object manager with this commit, too, otherwise I fear it will be forgotten as we consider the deprecation "done".

susannemoog (changes requested):
see comment above

linawolf:
I replaced the other usages.

susannemoog (approved):

PR #474, opened 2021-09-15
Fix unneccessary escaping of &amp; within literal URL

cweiske:


Documentation/7-Controllers/1-Creating-Controllers-and-Actions.rst
```diff
@@ -145,7 +145,7 @@ Usually, the display of a single Object is called by a link in the
 frontend. In this example extension it connects the list view by something
 like the following URL:
 
-`http://localhost/index.php?id=123&amp;tx_sjroffers_pi1[offer]=3&amp;tx_sjroffers_pi1[action]=show&amp;tx_sjroffers_pi1[controller]=Offer`
+``http://localhost/index.php?id=123&tx_sjroffers_pi1[offer]=3&tx_sjroffers_pi1[action]=show&tx_sjroffers_pi1[controller]=Offer``
```
mbrodala, line 148: What are the double backticks for? Inline code is normally wrapped with single backticks instead.

cweiske: In rST, double backticks are inline literal code: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-literals

Single backticks are "interpreted text", which is dependent on the renderer: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#interpreted-text

This is the reason github does not display single backticked strings with monospace font, but the rendered typo3 documentation does.

mbrodala: OK, but in this case we should indeed stick to single backticks which is preconfigured with the `:code:` role:

https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingReST/InlineCode.html#text-roles

cweiske: But I just see that a default role is defined in https://github.com/TYPO3-Documentation/TYPO3CMS-Book-ExtbaseFluid/blob/c611b1d52118cb98ada6ec0a9cc99b4d58c34aa9/Documentation/Includes.rst.txt

mbrodala (approved):

PR #487, opened 2021-11-20
Fix : undefined createHtmlResponse is called

fawmi:
The example calls the method `createHtmlResponse`, which is undefined. 
This PR replaces the `createHtmlResponse ` with `$this->htmlResponse();`


mbrodala (approved):

mbrodala:
Releases: master, 11.5

PR #515, opened 2022-03-08
no "use" of class domain allowed in file ext_localconf.php

franzholz:
And "use" cannot be used in file tt_content.php as well.


mbrodala:
Actually this is [fine since TYPO3v11](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.4/Important-94280-MoveContentsOfExtPhpIntoLocalScopes.html). Thus closing.

Issue #590, opened 2022-12-31
Direct people to newer versions of the content? (> v10.4)


sypets:
This manual is now only displayed for versions up to to 10.4. Can we have a way to direct people to newer versions of the content, e.g. have  a banner at the top.

I googled for something and landed on https://docs.typo3.org/m/typo3/book-extbasefluid/10.4/en-us/7-Controllers/1-Creating-Controllers-and-Actions.html. It would be great if there was some simple way to get to the documentation for newer TYPO3 versions.

sypets:
I noticed the problem again:

1. I was on the v10.4 page "[Content Elements & Plugins](https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/ContentElements/Introduction.html)" since it was the first page which the search results showed
2. clicked on link "[Configuring the Plugin](https://docs.typo3.org/m/typo3/book-extbasefluid/10.4/en-us/4-FirstExtension/7-configuring-the-plugin.html#configuring-the-plugin)" in the Extbase / Fluid Book
3. Would now like to switch to version I use (v11 or 12), is not possible

PR #591, opened 2023-03-21
Add notice with links to newer Extbase / Fluid content (10.4 only)

sypets:
Also, the Includes.rst.txt was not included on some pages. This was added as well.

The manual is currently only rendered up to 10.4. There is no version selector to switch to a newer version and there is no easy way to get to the newer version of a page unless one starts another search.

While a link on each page to link to newer content of this page would be preferable (from point of view of the reader), this is not feasible to do for every page.
However stating where the new content should be ("TYPO3 Explained") might be helpful when searching for it and adding a link to the respective chapters might be sufficient.

(Unfortunately, when googling one often ends up on older pages unless this is specifically prevented by specifying the versions).

In addition to this, a link to the newer content of the page might be helpful on some of the pages. Here, it might be a good idea to check which pages are viewed most frequently.

Resolves: #590


Documentation/Includes.rst.txt
```diff
@@ -32,3 +32,12 @@
...
 .. highlight:: php
+
+.. ---------------------
+.. Extbase moved message
+.. ---------------------
+
+.. attention::
+   This manual is no longer being maintained for TYPO3 versions 11.5 and above. The majority of the content has been
+   migrated to the :ref:`Extbase <t3coreapi_v11:extbase>` or :ref:`Fluid <t3coreapi_v11:fluid>` sections in
+   "TYPO3 Explained" or respective sections in the :ref:`TypoScript Reference <t3tsref_v11:start>`
```
brotkrueml, line 43: Which respective sections? FLUIDTEMPLATE?

DavidBruchmann: Import TypoScript settings are covered here already: https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ExtensionArchitecture/Extbase/Reference/TypoScriptConfiguration.html

sypets: @DavidBruchmann 
> Import TypoScript settings are covered here already: https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ExtensionArchitecture/Extbase/Reference/TypoScriptConfiguration.html

I thought we had moved some stuff to the TypoScript Reference. @linawolf would have better insight here.

linawolf: This is syntax, not reference stuff, so it is in typo3 explained

DavidBruchmann: @linawolf this distinction might not be so clear for users.

DavidBruchmann (approved):

